PR #5, opened 2020-07-19
Add docstrings and skeleton sphinx bits.

grahamgower:


demes/demes.py
```diff
@@ -55,6 +55,18 @@ def validator(self, attribute, value):
...
+    Population size parameters for a deme in a specified time period.
+    Times follow the backwards-in-time convection (time values increase
+    from the present towards the past).
+
+    :ivar start_time: The start time of the epoch.
+    :ivar end_time: The end time of the epoc.
+    :ivar initial_size: Population size at ``start_time``.
+    :ivar final_size: Population size at ``end_time``.
+        If ``initial_size != final_size``, the population sizes change
+        exponentially between the start and end times.
```
molpopgen, line 68: Do we want to say "exponential" here?  While that's the most common form of growth that people will want to model, one could imagine other functions, such as linear or logistic.  Should the "flavor" of growth be an adjective that is part of the class data?

grahamgower: You're right, this is stuck in a coalescent mindset. I guess a user-defined function provides the most generality here. I'll just change the language for now.

PR #16, opened 2020-07-20
Add comprehensive type annotations and stricter checks with mypy.

grahamgower:


grahamgower:
Closing this because the code has moved on. I might revisit this in the future.

Issue #7, opened 2020-07-19
Allow ancestor merge time to be flexible


apragsdale:
Right now, if I understand correctly, if a deme has an ancestor, it's end time has to be equal to the ancestors start time, so that ancestor and descendant populations cannot overlap. I think we want to relax this, and just check that the end time of a deme falls within the ancestors time interval.

As a very simple scenario, I'm imagining a demography that looks like the picture attached here, which has a main population and some offshoots. We don't want to have to set the three epochs along the main pop's branch as three separate demes.

From a user's perspective, I think we'd want the yaml description to look something like this. Then when reading it in, we'd validate that both end times of 500 and 1,000 are in [0, inf).

```YAML
description: An example demography with one main population that is unchanged, that
  spawns multiple offshoot populations.
time_units: generations
demes:
  - ancestral:
    description: Main population
    start_time: 0
    initial_size: 1000
  - offshoot1:
    description: More recent offshoot population
    ancestors: ancestral
    start_time: 0
    end_time: 500
    initial_size: 100
  - offshoot2:
    description: More ancient offshoot population
    ancestors: ancestral
    start_time: 0
    end_time: 1000
    initial_size: 200
```

![IMG_20200717_182949](https://user-images.githubusercontent.com/33734266/87886040-f4ce2380-c9e7-11ea-8b86-de2a9c64e42d.jpg)

grahamgower:
I added the `DemeGraph.subgraph()` method for this reason. The idea is that you add a new deme to the graph which has no `ancestor`, but is attached via pulse migrations. The `browning_america()` example model uses it to do multiple ancestors, but there's no reason it couldn't be used for your example too. Do you think that's sufficient? https://github.com/grahamgower/demes/blob/6ae9c9c1f521a33e9c22d632ca216a44a21c260a/demes/examples.py#L48-L74

Conceptually, it seems fine to allow attachment to an ancestor at any time of its existence, as you suggest. The behaviour of implicitly setting a deme's `end_time` to its ancestor's `start_time` should probably be removed if we allow this though. I think having the implict behaviour, and also supporting the override, could lead to confusion.

apragsdale:
Ok, great. I'll play around with that a bit more! I think this would work. So if we make those changes, we would be able to define that demography just as in that example I attached, and internally handle it with mass migration events?

grahamgower:
[...]

PR #17, opened 2020-07-21
Start features for forward time

apragsdale:
This is a start for changing the flow of time to forward instead of backward, and adding some vocabulary for the different demographic events, such as splits, branches, mergers, and admixtures, as proposed in #7. There is still quite a bit to play around with here, but I wanted to get at least an unpolished PR up with some of those ideas so there's a reference for discussion. I'll fill it out a bit more tomorrow as well.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -63,7 +63,7 @@ class Epoch:
     start time must be greater than the end time
 
     :ivar start_time: The start time of the epoch.
-    :ivar end_time: The end time of the epoch.
+    :ivar end_time: The end time of the epoch (must be specified).
     :ivar initial_size: Population size at ``start_time``.
     :ivar final_size: Population size at ``end_time``.
         If ``initial_size != final_size``, the population size changes
@@ -117,7 +117,7 @@ class Migration:
     """
     source: ID = attr.ib()
     dest: ID = attr.ib()
-    start_time: Time = attr.ib(validator=[non_negative, finite])
+    start_time: Time = attr.ib(validator=[non_negative])
     end_time: Time = attr.ib(validator=[non_negative, finite])
     rate: Rate = attr.ib(validator=[non_negative, finite])
 
@@ -843,9 +843,9 @@ class DemeGraph:
                                     dict(demes=[source, dest], rate=m["rate"])
                                 )
                                 if "start_time" in m:
-                                    m_asymmetric[-1]["start_time"] = m["start_time"]
+                                    m_symmetric[-1]["start_time"] = m["start_time"]
                                 if "end_time" in m:
-                                    m_asymmetric[-1]["end_time"] = m["end_time"]
+                                    m_symmetric[-1]["end_time"] = m["end_time"]
                                 # pop the m_compare so we don't repeat it
                                 m_dict[(dest, source)].remove(m_compare)
                                 no_symmetry = False
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -63,20 +63,20 @@ class Epoch:
     start time must be greater than the end time
 
     :ivar start_time: The start time of the epoch.
-    :ivar end_time: The end time of the epoc.
+    :ivar end_time: The end time of the epoch.
     :ivar initial_size: Population size at ``start_time``.
     :ivar final_size: Population size at ``end_time``.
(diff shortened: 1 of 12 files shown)


grahamgower:
Hmmm.. I wonder if it might be cleaner/easier to just throw away most of the existing `demes.py` code.

demes/demes.py
```diff
@@ -143,21 +149,113 @@ def __attrs_post_init__(self):
             raise ValueError("source and dest cannot be the same deme")
 
 
+@attr.s(auto_attribs=True)
+class Split:
+    """
+    Parameters for a split event, in which a deme ends at a given time and
+    contributes ancestry to an arbitrary number of descendant demes. Note
+    that there could be just a single descendant deme, in which case ``split``
```
jeromekelleher, line 157: We could enforce > 1 child deme?

apragsdale:
> Hmmm.. I wonder if it might be cleaner/easier to just throw away most of the existing `demes.py` code.

No, I think it's a great start. All the changes here are just ideas I'm trying out and it doesn't mean we need to stick with them exactly or throw out the backbone of what's in there now.

Some of my changes are pretty clunky, like the `ancestor` vs `ancestors` option. We probably want to wrap those into one keyword. Right now it's set up to read in a yaml demography (`g = demes.load("browning_america.yaml")`, e.g.), and then determine the demographic events and add them to the event lists by calling `g.finalize_events()` (though maybe that function should be called `fill_events` or `get_events`). So aside from `migrations` and `pulses` between existing demes described in the yaml file, the other events are implicit and determined from the predecessor/successor relationships of the demes.

demes/demes.py
```diff
@@ -57,8 +57,10 @@ def validator(self, attribute, value):
 class Epoch:
     """
     Population size parameters for a deme in a specified time period.
-    Times follow the backwards-in-time convention (time values increase
-    from the present towards the past).
+    Times follow the forwards-in-time convention (time values increase
+    from the present towards the past). The start time of the epoch is
+    the more ancient time, and the end time is more recent, so that the
+    start time must be greater than the end time
```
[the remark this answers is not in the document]
apragsdale: That is a bit confusing, but I wanted to avoid having to write negative numbers. Not sure what the best solution for anchoring and measuring relative time is.

demes/demes.py
apragsdale: I think that makes sense for a "split". Should we enforce that if two demes are defined (parent/child), and no other demes are involved (i.e., just a "hand-off" from one deme to the next), that users need to define that as a single deme?

apragsdale:
I added/reworked a bit more here to try to write out symmetric and asymmetric migrations, and there are some examples I pushed to an examples directory that I've been testing with. Very interested to hear what you think! This stuff is pretty fun to play with :)

molpopgen:
Is `proportions` a sufficient descriptor, or would something more verbose like `ancestry_proportions` be more clear?

From the Browning model output:

```py
    admixtures=[
        Admix(
            parents=["AFR", "EUR", "EAS"],
            proportions=[0.167, 0.333, 0.5],
            child="ADMIX",
            time=12,
        )
    ],
```

grahamgower:
This looks really good @apragsdale! The examples look pretty clear to me. I think our next move should be to decide on variable/keyword naming for time, and a direction convention.

>Is proportions a sufficient descriptor, or would something more verbose like ancestry_proportions be more clear?

I'm on the fence here. Generally, I prefer shorter keywords if they suffice. But the number of times a user encounters this keyword will be small, so it might be sensible to be more explicit here.

grahamgower:
Hey @apragsdale, I think we should get this merged ASAP. Do you want to [`@unittest.skip`](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) the failing tests for now and merge?

apragsdale:
Yes, let's merge this in soon. Let me take a quick pass later today through the current set of tests, because I haven't touched them, since I wasn't sure if this was going to be the direction we go. Later on this week I can develop a more full suite of tests that can be pulled in at a later point, how about?

grahamgower:
This looks great, thanks @apragsdale! Can you merge it if you're happy?

PR #18, opened 2020-07-21
The two most important goals were missing.

grahamgower:


README
```diff
@@ -6,6 +6,8 @@ how they change over time, and their relationships to one another.
 of a demographic model.
 
 ## Goals
+- A simple declarative format for specifying a demographic model.
+- Unify the specification of demographic models across different frameworks.
```
jeromekelleher, line 10: I would say something like "Portable and unambiguous description of demographic models that can be used by multiple simulation frameworks".

This won't help for the target audience, but I'm thinking it's something like the high-level description is like Java source code, and the low-level description is like byte code. So, we're compiling the human readable description into a machine readable description made up of low-level primitive operations.

grahamgower: Oops, I already merged. Would you like to just go ahead and make any changes you see fit?

jeromekelleher: OK, will do.

PR #28, opened 2020-08-15
Add to/from msprime/stdpopsim conversion functions.

grahamgower:
I expect demes->msprime will be one of the shortest export functions we have. I didn't really test this, other than eyballing the demography debugger output below. So, use at your own risk!

```
$ python demes2msprime.py ../examples/gutenkunst_ooa.yml
Model =  hudson(reference_size=1)
===============================
Epoch: 0 -- 21200.0 generations
===============================
     start     end      growth_rate |     0        1        2        3        4        5
   -------- --------       -------- | -------- -------- -------- -------- -------- --------
0 | 7.3e+03  7.3e+03              0 |     0        0        0        0        0        0
1 |1.23e+04 1.23e+04              0 |     0        0        0        0        0        0
2 | 2.1e+03  2.1e+03              0 |     0        0        0        0        0        0
3 |1.23e+04 1.23e+04              0 |     0        0        0        0      3e-05   1.9e-05
4 |2.97e+04   1e+03         0.00016 |     0        0        0      3e-05      0     9.6e-05
5 |5.41e+04    510          0.00022 |     0        0        0     1.9e-05  9.6e-05     0

Events @ generation 21200.0
   - Population parameter change for 2: initial_size -> 2100 growth_rate -> 0
   - Mass migration: Lineages moved with probability 1.0 backwards in time with source 4 & dest 2
                     (equivalent to migration from 2 to 4 forwards in time)
   - Mass migration: Lineages moved with probability 1.0 backwards in time with source 5 & dest 2
                     (equivalent to migration from 2 to 5 forwards in time)
   - Migration rate change for (2, 3) to 0.00025
   - Migration rate change for (3, 2) to 0.00025
   - Migration rate change for (4, 3) to 0
   - Migration rate change for (3, 4) to 0
   - Migration rate change for (5, 3) to 0
   - Migration rate change for (3, 5) to 0
   - Migration rate change for (5, 4) to 0
   - Migration rate change for (4, 5) to 0
======================================
Epoch: 21200.0 -- 140000.0 generations
======================================
     start     end      growth_rate |     0        1        2        3        4        5
   -------- --------       -------- | -------- -------- -------- -------- -------- --------
0 | 7.3e+03  7.3e+03              0 |     0        0        0        0        0        0
1 |1.23e+04 1.23e+04              0 |     0        0        0        0        0        0
2 | 2.1e+03  2.1e+03              0 |     0        0        0     0.00025     0        0
[...]

PR #32, opened 2020-08-30
Add DemeGraph.in_generations(), which converts to units of generations.

grahamgower:
I've separated this out from the other pr (#28), so we can consider the semantics in isolation.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -2,6 +2,7 @@ from typing import List, Mapping, Union
 import itertools
 import math
 import collections
+import copy
 
 import attr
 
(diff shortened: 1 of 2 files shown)


demes/demes.py
```diff
@@ -388,10 +389,11 @@ class DemeGraph:
     :ivar description: A human readable description of the demography.
     :ivar time_units: The units of time used for the demography. This is
         commonly ``years`` or ``generations``, but can be any string.
-    :ivar generation_time: The generation time of demes.
-        TODO: The units of generation_time are undefined if
-        ``time_units="generations"``, so we likely need an additional
-        ``generation_time_units`` attribute.
+        This field is intended to be useful for documenting a demography,
```
jeromekelleher, line 392: It took me a few minutes to see that the units cancel out here, and it all works once your time values are all the same units (including generation_time). We should probably document this a bit more clearly.

jeromekelleher (commented):
LGTM - as I say, took me a while to see that ``time_units`` doesn't really do anything so maybe good to explain that a bit more slowly.

PR #38, opened 2020-10-19
adjust migration rate in in_generations

### LINKED ISSUE #37: in_generations() is not rescaling migration rates
Looks like the `in_generations()` function does not rescale migration rates (i.e. converting from years to generations does not change the migration rates from units of per-year to per-generation).

Should be a quick fix - I'll have a go at it (unless this was the intended behavior?)

apragsdale:
Fixes #37


apragsdale:
Gonna close - previously agreed on the choice to have all rates in units of per generation, even if times are given in years or other units.

PR #64, opened 2020-10-31
Migration validity checks

apragsdale:
We don't currently check that migrations and pulses occur only at times that both demes involved exist. This just ensures that migration epochs don't extend outside the time intervals of demes involved, and pulse times occur when both demes are present as well.

Now if there is an invalid migration or pulse event, we raise a ValueError.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -8,6 +8,8 @@ import copy
 import attr
 from attr.validators import optional
 
+from .script import dumps, loads
+
 Number = Union[int, float]
 ID = str
 Time = Number
@@ -908,6 +910,12 @@ class DemeGraph:
                     )
         for deme_from, demes_to in splits_to_add.items():
             self.split(deme_from, list(demes_to), self[deme_from].end_time)
+
+    def validate(self):
+        """
+        Validates the demographic model.
+        """
+        loads(dumps(self))
 
     def in_generations(self):
         """
--- a/tests/test_demes.py
+++ b/tests/test_demes.py
@@ -15,6 +15,8 @@ from demes import (
     load,
 )
 
+import demes
+
 
 class TestEpoch(unittest.TestCase):
     def test_bad_time(self):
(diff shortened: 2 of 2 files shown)


grahamgower:
Aren't we already checking this with calls to `DemeGraph.check_time_intersection()` within `DemeGraph.pulse()` and `DemeGraph.migration()`?

apragsdale:
Hmm, you're right. I don't know why I missed that. In the test, I realize that I implemented migration by just creating a list of `demes.Migration` objects as in `dg.migrations = [demes.Migration(...)]`, instead of calling `g.migration(...)` which would have run those checks.

So looks like I ended up confusing myself by implementing those test demographies in a roundabout way. Does that suggest though that we want to move where we run those checks to make sure invalid demographies don't slip through?

grahamgower:
I suspect that in Python it will always be possible to bypass any validation checks we put in. Though it might be worth having an explicit "validate" function, for cases where someone really wants to bypass the api (e.g. to temporarily create an invalid demography, and then fix it up afterwards), but then check for consistency later. We could implement such a validation function as a one-liner that does `demes.loads(demes.dumps(deme_graph))` for now.

demes/demes.py
```diff
@@ -909,6 +911,12 @@ def get_demographic_events(self):
         for deme_from, demes_to in splits_to_add.items():
             self.split(deme_from, list(demes_to), self[deme_from].end_time)
 
+    def validate_demography(self):
```
grahamgower, line 914: Maybe we could call this `validate()`? I think it's clear the demography is going to be validated because it's a method of the DemeGraph.

grahamgower (approved):
Looks good, thanks @apragsdale!

demes/demes.py
apragsdale: Sounds good! Will change that and merge.

PR #66, opened 2020-11-11
Paper outline

apragsdale:
Thought I'd get a draft/outline of an eventual paper up here. Hope this isn't way too premature! :)


molpopgen:
Makes sense, and this prevents @apragsdale and I from suggesting we do it in R markdown. 🤣

grahamgower:
Wow! This looks really good, thanks @apragsdale! Do we want to include this here though, or have a separate repository for the manuscript?

jeromekelleher:
Yes, this is *awesome* thanks @apragsdale! But I second the idea of spinning this into its own repo. There's no point in weighing down the software repo here with latex etc, and it'll make it easier to keep track of issues on the paper if we can use the issue tracker for that. Small focused repos are good.

apragsdale:
Ok, will do! That makes perfect sense to me. I'll make a demes paper repo and share.

Issue #68, opened 2020-11-11
improve test coverage


grahamgower:
In particular, there are a bunch of code paths in `demes/demes.py` that we should try to hit.

apragsdale:
Yes, good call. I'll take a pass at this this afternoon. Should we set up a Codecov check for this repo, as well?

grahamgower:
I'm going to reopen this, because I think there's still a bunch left to do. I'll work on it a bit today too. Agreed that we should set up Codecov.

Issue #33, opened 2020-10-06
add/fix method for DemeGraph equality


grahamgower:
We want to be able to test for equality between two deme graphs, e.g. for the test suite. Being an `attrs` class, `demes.DemeGraph` has an `__eq__` method by default. It does the right thing for most attrs classes, and this is the code attrs generates right now:
```python
>>> import demes, inspect
>>> print(inspect.getsource(demes.DemeGraph.__eq__))
def __eq__(self, other):
    if other.__class__ is not self.__class__:
        return NotImplemented
    return  (
        self.description,
        self.time_units,
        self.generation_time,
        self.default_Ne,
        self.doi,
        self.demes,
        self.migrations,
        self.pulses,
        self.splits,
        self.branches,
        self.mergers,
        self.admixtures,
        self.selfing_rate,
        self.cloning_rate,
    ) == (
        other.description,
        other.time_units,
        other.generation_time,
        other.default_Ne,
        other.doi,
        other.demes,
        other.migrations,
        other.pulses,
        other.splits,
        other.branches,
        other.mergers,
        other.admixtures,
        other.selfing_rate,
        other.cloning_rate,
    )
```

But in our case we likely want to ignore the ordering of items in the `.demes`, `.pulses`, and `.migrations` lists. One way of dealing with this is to enforce some ordering for these lists inside the DemeGraph. But this has undesirable consequences: (1) round-tripping of yaml files would give different output ordering, and (2) the order used by a yaml script writer probably has some useful semantic meaning (such as ordering demes starting from the root, keeping all migrations for deme A together, etc.).

We could manually define an `__eq__` such that some attributes (or all `list` attributes) get sorted, like:
```python
def __eq__(self, other):
    if other.__class__ is not self.__class__:
        return NotImplemented
    return  (
        sorted(self.demes),
        sorted(self.migrations),
        sorted(self.pulses),
        ...
    ) == (
        sorted(other.demes),
        sorted(other.migrations),
        sorted(other.pulses),
        ...
    )
```
But then we'd need to be careful to update the method if/when we add or remove attributes or otherwise change the behaviour. It would be nice to instead do some [magic code generation using `__attrs__attrs__`](https://www.attrs.org/en/stable/extending.html)

grahamgower:
And floating point numbers should be compared with `math.isclose()`. Attrs has an open pr to [support passing a custom comparison function](https://github.com/python-attrs/attrs/pull/627) that would allow us to do:
```python
@attr.s
[...]

PR #72, opened 2020-11-12
add equality tests

grahamgower:
Part of #68.

I've made a decision about equality semantics here. Some attributes of the deme graph are just used as defaults for setting attributes on added demes (`default_Ne`, `selfing_rate`, `cloning_rate`), and I consider this a model implementation detail. So here we only check if the demes in the graphs compare equal, rather than caring about the values of these convenience attributes.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -4,6 +4,7 @@ import itertools
 import math
 import collections
 import copy
+import operator
 
 import attr
 from attr.validators import optional
@@ -17,6 +18,8 @@ Size = Number
 Rate = float
 Proportion = float
 
+_ISCLOSE_REL_TOL = 1e-9
+_ISCLOSE_ABS_TOL = 1e-12
 
 # Validator functions.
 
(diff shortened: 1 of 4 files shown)


demes/demes.py
```diff
@@ -92,7 +92,10 @@ def isclose(a: Optional[Number], b: Optional[Number]) -> bool:
         return (a,) == (b,)
     else:
         return math.isclose(
-            typing.cast(typing.SupportsFloat, a), typing.cast(typing.SupportsFloat, b)
+            typing.cast(typing.SupportsFloat, a),
+            typing.cast(typing.SupportsFloat, b),
+            rel_tol=1e-9,
```
jeromekelleher, line 97: Always tricky deciding what these tolerances should be - why have you changed them here? I think it'd be better to add the parameters to the function using the same defaults as the builtin, and then change the tolerances in the specific cases where you need it. There's nasty gotchas down the line otherwise.

jeromekelleher (commented):
I don't think it's a good idea to have ``==`` correspond to a non-exact equality with some arbitrary tolerances that can't be controlled baked in. I think this could have really nasty effects down the line.

I think it would be better to have ``==`` correspond to exact byte-for-byte equality and to define a method ``equals(other, abs_tol, rel_tol)`` which does the comparison that we actually want. We could even call it ``isclose`` just to make the connection between the library function totally explicit.

Equality semantics is really important here...

demes/demes.py
grahamgower: [`math.isclose()`](https://docs.python.org/3/library/math.html#math.isclose) has defaults of `rel_tol=1e-9, abs_tol=0`. The rel_tol is fine, I just added that for clarity. I changed from the default `abs_tol=0` to `abs_tol=1e-12`, because I think we generally want `isclose(0, epsilon)` to be true.

grahamgower:
> I don't think it's a good idea to have `==` correspond to a non-exact equality with some arbitrary tolerances that can't be controlled baked in. I think this could have really nasty effects down the line.
> 
> I think it would be better to have `==` correspond to exact byte-for-byte equality and to define a method `equals(other, abs_tol, rel_tol)` which does the comparison that we actually want. We could even call it `isclose` just to make the connection between the library function totally explicit.

Thanks, that's a very nice suggestion. This probably means giving each class an `isclose` method, but that should be fine.

> Equality semantics is really important here...

Agreed!

grahamgower:
Well that took longer than expected, but I think the result is much nicer. It cleans up #33 so that we don't need any special features of `attr`.

jeromekelleher (approved):
Nice!

PR #87, opened 2020-11-15
update and rearrange API docs

grahamgower:
This adds some quickstart/examples at the top of the API page (using jupyter-sphinx). It moves the yaml schema to its own file to facilitate verbatim inclusion in the docs. The format of the schema isn't ideal for readers, but I didn't want to just duplicate it in a different format either. Maybe we could autogenerate something from this which is formatted more neatly?

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -616,24 +616,27 @@ class Deme:
 @attr.s(auto_attribs=True)
 class DemeGraph:
     """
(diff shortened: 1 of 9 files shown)


jeromekelleher:
It's a pity that strictyaml is so closely tied to Python and there's no text description of the schema. There's not much that can be done about that, though, I guess.

jeromekelleher:
Docs LGTM

PR #95, opened 2020-11-16
use keyword args almost everywhere

### LINKED ISSUE #92: use more keyword-only arguments
While keyword-only args can sometimes be annoying (it's more typing), it makes it impossible to provide arguments in the wrong order. It's also produces much easier to read code (such as tests), as the reader doesn't have to know the arg order or look up the arg order.

So I think we should change to `kw_only=True` for the attr's classes. And the methods on `DemeGraph` should be mostly keyword-only.

grahamgower:
Closes #92.


jeromekelleher (approved):
LGTM

Issue #97, opened 2020-11-17
DemeGraph methods: split(), branch(), admix(), and merge()


grahamgower:
[EDIT: fat fingers... I somehow submitted the issue before writing it]

The `DemeGraph` has `split()`, `branch()`, `admix()`, and `merge()`, methods, but the semantics of how these could be used are not clear. For example, we have the `split()` method.
```python
    def split(self, *, parent, children, time):
        """
        Add split event at a given time. Split events involve a parental deme
        whose end time equals the start time of all children demes.

        :param str parent: The ID of the parent deme.
        :param children: A list of IDs of the descendant demes.
        :type children: list of str
        :param float time: The time at which split occurs.
        """
        for child in children:
            # check parent/children relationship and end/start times
            if child == parent:
                raise ValueError("cannot be ancestor of own deme")
            if self[parent].end_time != self[child].start_time:
                raise ValueError(
                    f"{parent} and {child} must have matching end and start times"
                )
            # the ancestor of each child population is set
            self[child].ancestors = [parent]
        self.splits.append(Split(parent=parent, children=children, time=time))
```
This suggests that the `parent` and `children` must already exist in the deme graph (a ValueError will be raised for non-existent children in `self[child]`). So the children and the parent were already added with calls to `deme()`, but the ancestors field wasn't used? To me, these methods feel redundant at best, and confusing at worst.

Presumably, these methods should be internal to `get_demographic_events()` and not exposed in the Python API?

jeromekelleher:
I don't have a good feel for how the Python API is supposed to be used right now so :shrug: Hopefully I'll be more helpful once I've done the ms converter and got my hands dirty a bit.

apragsdale:
> Presumably, these methods should be internal to `get_demographic_events()` and not exposed in the Python API?

That's right, I think. The way things currently work, is you could 1) define demes and include the `ancestors` for the child demes for a split event, e.g., or 2) define the demes without setting ancestors and then call `split`, e.g., to set the ancestors and check that the time intersection is correct.

Some of the events, such as `merge`, also have the effect of truncating parental demes to end at the time of the merger event, so you could do something like:
```
import demes
g = demes.DemeGraph(description="test", time_units="generations")
g.deme(id="parent1", initial_size=1000)
g.deme(id="parent2", initial_size=1000)
g.deme(id="child", initial_size=1000, start_time=10)
g.merge(parents=["parent1", "parent2"], child="child", proportions=[0.5, 0.5], time=10)
```
[...]

Issue #96, opened 2020-11-17
DemeGraph.get_demographic_events()


grahamgower:
This method classifies ancestor/descendant relationships into `split`/`branch`/`admix`/`merge` types. These classifications are currently used by the `moments` conversion code, but might be more widely useful. However, `get_demographic_events()` only gets called on a deme graph when a yaml file is loaded, in `demes.loads()`. So if one wishes to instead use the python API to construct a graph, then one must first call `get_demographic_events()` onself before using the `SFS()` moments conversion fuction. Perhaps `demes.convert.SFS()` should be calling this method itself? I also wonder if `get_demographic_events()` should instead return the splits/branches/admixtures/mergers lists instead of these being attributes of the graph? And maybe we could rename the method? `classify_graph_vertices()`?

apragsdale:
That could make sense, or perhaps `list_demographic_events` if we change the behavior to return lists of all discrete demographic events. Or even just `g.demographic_events()` returns a dictionary of event types as keys and lists of events as values.

I'm realizing now, though, that the behavior is wrong. Currently, if I call `g.get_demographic_events()` more than once, events get added multiple times to those lists instead of being cleared before populating them. Or if an event is already in one of the lists, it gets added again. That would need to be fixed if we don't change the behavior here.

apragsdale:
Closed by #110

PR #101, opened 2020-11-17
Remove demographic event attributes

apragsdale:
Addresses #96, #97. The `split`, `merge`, etc functions in the `DemeGraph` class now are underscored. `get_demographic_events()` has been changed to `list_demographic_events()` and returns a dictionary of lists of discrete demographic events.

Is there a better name for `list_demographic_events` or is that fine?

### DIFF
--- a/demes/convert/moments_.py
+++ b/demes/convert/moments_.py
@@ -51,7 +51,7 @@ def SFS(g, sampled_demes, sample_sizes, sample_times=None, Ne=None, unsampled_n=
         sample_times = [g[d].end_time for d in sampled_demes]
 
     # for any ancient samples, we need to add frozen branches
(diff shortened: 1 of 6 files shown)


apragsdale:
I _think_ this is ok, but definitely could use an extra set of eyes or suggestions for API tweaks if needed @grahamgower

grahamgower:
Looks good @apragsdale, thanks! I'm not sure about the name `list_demographic_models()` either. But to be honest, I don't have a better suggestion. I still have some questions about the newly underscored methods, but I'll take that over to #97.

apragsdale:
Yeah, I'm not sure of a better name either. I'll merge what we have here for now, and if we want to update the name later we can come back to it.

grahamgower:
Sounds good!

PR #102, opened 2020-11-17
Initial ms parser

jeromekelleher:
Very much a WIP, but I thought I'd share just so you know I'm working on it at least!

The basic approach of adapting the msprime argparse code will work I think, and I think the error handling for parsing is working well - setting ``parser.error = ValueError`` seems to do the trick quite nicely.

One high level question though: what should we do with ms command line args that aren't to do with directly defining the demography? There's two classes I think:

1. Arguments for defining the sample size. This is intrinsically linked up with defining the demography, so I'm currently returning a tuple ``DemeGraph, num_samples``, so that we return the parsed-out number of samples per deme. I'm not at all sure this is the right approach - maybe we should just ignore all the sample size stuff?

2. What do we do with arguments like ``num_replicates``, ``-T``, ``-t`` etc? Should we accept an valid ms command string, but ignore any stuff that isn't about defining demography? Currently we'll error on any arguments that aren't about the demography, but I'm not sure this is the correct thing.

Anyway, thoughts appreciated. i wouldn't bother looking at the code much, it's not ready for review.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,3 +58,9 @@
 .. autoclass:: demes.Admix
     :members:
 ```
+
+## Conversion functions
+
+```{eval-rst}
+.. autofunction:: demes.from_ms
+```
--- a/demes/__init__.py
+++ b/demes/__init__.py
@@ -21,3 +21,4 @@ from .demes import (
     Admix,
 )
 from .load_dump import load_asdict, loads_asdict, load, loads, dump, dumps
+from .ms import from_ms
--- a/demes/ms.py
+++ b/demes/ms.py
@@ -1,0 +1,842 @@
+import copy
+import math
+import argparse
+import logging
+import sys
+from typing import Any, Dict, List, Mapping, MutableMapping, Set, Tuple
+
+import attr
+
+import demes
+from .demes import finite, positive, non_negative, unit_interval
+
+logger = logging.getLogger(__name__)
+
+
+class ValueErrorArgumentParser(argparse.ArgumentParser):
+    """
+    An AgumentParser that throws a ValueError instead of exiting.
+    It would be preferred to catch and reraise TypeError and ValueError
+    separately, but that isn't possible without touching ArgumentParser
+    internals because those errors are caught by parse_known_args(),
+    which then throws an argparse.ArgumentError.
+    So we just use ValueError for everything.
+    """
+
+    def error(self, message):
+        _, exc, traceback = sys.exc_info()
+        raise ValueError(str(exc))
+
+
+def coerce_nargs(obj_creator, append=False):
+    """
+    Return an argparse action that coerces the collected nargs into
+    data class using the specified class/function ``obj_creator``.
+    This simplifies the validation of heterogeneously-typed args lists.
+    If ``append`` is True, the option may be specified multiple times
+    and each coerced object will be appended to a list.
+
+    .. code::
+
+        @attr.define
+        class Foo
+            n = attr.ib(converter=int)
+            x = attr.ib(converter=float)
+
+        parser.add_argument(
+            "--foo",
+            nargs=2,
+            action=coerce_nargs(Foo),
+            ...
+        )
+
+        args = parser.parse_args("--foo 6 1e-5")
+        assert isinstance(args.foo, Foo)
+
+    """
+    parent_class = argparse.Action
+    if append:
+        parent_class = argparse._AppendAction
+
+    class CoerceAction(parent_class):
+        def __call__(self, parser, namespace, values, *args, **kwargs):
+            obj = obj_creator(*values)
+            # We save the option strings to later identify which option was
+            # used to create the object. E.g. -n and -en options both map to
+            # PopulationSizeChange, but have slightly different semantics.
+            obj.option_strings = self.option_strings
+            if append:
+                super().__call__(parser, namespace, obj, *args, **kwargs)
+            else:
+                setattr(namespace, self.dest, obj)
+
+    return CoerceAction
+
+
+##
+# Data classes for ms options. These help to separate input validation
+# from the graph building procedure.
+
+
+@attr.define
+class Option:
+    # This attribute is set by CoerceAction.
+    option_strings: List[str] = attr.ib(init=False)
+
+
+@attr.define
+class Structure(Option):
+    # -I npop n1 n2 ... [4*N0*m]
+    npop = attr.ib(converter=int, validator=positive)
+    n: Any = attr.ib()  # samples list: currently ignored
+    rate = attr.ib(converter=float, validator=non_negative)
+
+    def __attrs_post_init__(self):
+        if len(self.n) != self.npop:
+            raise ValueError("sample configuration doesn't match number of demes")
+
+    @classmethod
+    def from_nargs(cls, *args):
+        npop, *n = args
+        rate = 0
+        try:
+            npop = int(npop)
+        except ValueError:
+            raise ValueError(f"-I 'npop' ({args[0]}) not an integer")
+        if len(n) == npop + 1:
+            *n, rate = n
+        return cls(npop, n, rate)
+
+
+@attr.define
+class Event(Option):
+    # The first param of every demographic event is the event time.
+    # For ms options without a time parameter (i.e. those that set the
+    # initial simulation state), t is just set to zero.
+    t = attr.ib(converter=float, validator=non_negative)
+
+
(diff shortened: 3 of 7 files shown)


jeromekelleher:
Edit: I should add that the plan is *not* to import msprime like it's doing at the moment

grahamgower:
Cool! Thanks @jeromekelleher! In general, I think we could just warn about arguments that are unused/untranslated. Maybe have a `warn_unused` param to `from_ms()` that defaults to True? Returning the sample info is a good idea too. But probably we need to think harder about sampling?

It's clear that all simulators that consume deme graphs will also need to think about samples. At least, users will specify sampling *somehow*, and the samples and sampling times should at least be checked against the demes and deme existence times. I think it will also be common to want to `prune` a deme graph, given a set of samples, to just the ancestry of those samples.

jeromekelleher:
> It's clear that all simulators that consume deme graphs will also need to think about samples. At least, users will specify sampling somehow, and the samples and sampling times should at least be checked against the demes and deme existence times. 

Yes, I guess we do need some way of describing samples, although I'd hoped that this could be left to the individual simulators. In principle, forward simulators don't need samples.

> I think it will also be common to want to prune a deme graph, given a set of samples, to just the ancestry of those samples.

Yes, good idea. So we reason about the demes that are unreachable from the sampling demes, given the migration rates etc. Sounds do-able.

jeromekelleher:
I'm not making a whole pile of progress on this, and it seems I'm unlikely to get the time to actually get into this for while. If anyone else would like to pick this up, I'd be happy to give pointers on where to go with it. Basically, the text parsing is taken care of and we "just" need to translate the model into a demes graph. 

Just in case anyone is wondering where this PR is going ...

grahamgower:
I've just pushed a commit that completes the ms converter. The API is open to discussion, and test converage is still low, but it works for the examples in the ms manual and for the zigzag and Gutenkunst OOA models.

demes/convert/ms.py
```diff
@@ -0,0 +1,757 @@
...
+        and ms model, trying setting this flag to ``True``.
+
+        Ms permits migration
+        into a deme at times before the deme splits (after an ``-ej``
+        option is provided). In Demes, this is explicitly disallowed,
+        and such models cannot be converted to a demes graph. The correct thing
+        to do when using ms is to explicitly disable migrations into the
+        deme immediately past-wards of the ``-ej`` option. By setting
+        ``disable_migration_before_split`` to ``True``, this will be
+        done for automatically, so that the model will be "fixed" and
```
jeromekelleher, line 735: for you

docs/api.md
```diff
@@ -58,3 +58,9 @@
 .. autoclass:: demes.Admix
     :members:
 ```
+
+## Conversion functions
+
+```{eval-rst}
+.. autofunction:: demes.convert.from_ms
```
jeromekelleher, line 65: I'm kinda anti hierarchy, unless it's really needed. ``demes.from_ms`` would be more convenient, wouldn't it? Particularly with the ``from_x`` form, I really can't see us having namespace collisions.

jeromekelleher (commented):
Looks great! I think if we should merge once the test coverage is bumped up - best way to do that is probably case-by-case with small examples.

Once msprime 1 is out, we could explore using some internal msprime APIs to beef up test coverage, comparing a bunch of command lines to the Demography object msprime creates. Might not be worth it though.

docs/api.md
grahamgower: Sure, that works.

demes/convert/ms.py
```diff
@@ -0,0 +1,757 @@
...
+
+    Ms uses coalescent units for times, population sizes, and migration rates.
+    These will be converted to more familiar units using the :math:`N_0` value
+    provided, according to the following rules:
+
+    .. math::
+
+        \\text{time} = 4 \\times N_0 \\times t
+
+        \\text{deme size} = N_0 \\times \\text{size}
```
grahamgower, line 709: I'm not so sure about this one.

demes/convert/ms.py
```diff
@@ -0,0 +1,757 @@
...
+        type=float,
+        default=[],
+        metavar=("t", "i", "j"),
+        help=(
+            "Move all lineages in population i to j at time t. "
+            "Forwards in time, this corresponds to a population split "
+            "in which lineages in j split into i. All migration "
+            "rates for population i are set to zero."
+        ),
+    )
```
grahamgower, line 295: So, this `-ej` option has me stumped. Firstly, the statement here that "All migration rates for population i are set to zero" is wrong. The ms manual says
> -ej t i j
Move all lineages in subpopulation i to subpopulation j at
time t. Migration rates from subpopulation i (M[k,i], k ≠ i) are set to
zero. (Prior to May 19, 2007, ms left migration rates unchanged .) In
the forward direction this corresponds to population splitting. Population
growth rates are unchanged.

The change in 2007 is to turn off migration in one direction only. I checked the code, and this seems to be exactly what the code is doing (the ms code is fairly difficult to grok, so maybe I missed something). But I can't see why one would care about that direction. With the post-2007 behaviour, lineages are no longer allowed to migrate (backwards in time) from population `i`. But after the event, there won't be any lineages in that population anyway, so who cares? However, lineages *can* migrate (backwards in time) into population`i`, causing almost the same issue that plagued the Gutenkust et al. OOA model example in the msprime manual.

I didn't see anything in [Ragsdale et al. 2020](https://doi.org/10.1016/j.ajhg.2020.08.017) that suggested the original ms implementation of the Gutenkunst et al. OOA model suffered from this problem. So either (1) I'm wrong, or (2) the ms code change in 2007 is wrong, and the indices should be reversed.

grahamgower: Perhaps foolishly, I assumed that the ms manual describes migration matrix indices and migration direction like msprime, where the source/dest of migration are with respect to lineages moving as the coalescent process progresses backwards through time. Indeed the ms manual states:

> ms will also produces samples under models with arbitrary backward mi-
gration matrices. This is done by first using the -I option and its arguments
(with or without the migration rate parameter.) This establishes the number
of subpopulations, the sample configuration, and assigns preliminary values to
the elements of the migration matrix. The elements of the migration matrix
are M[i][j] = 4*N0*m[i][j], i, j = 1, . . . ,npop , where m[i][j] is the fraction of subpopu-
lation i which is made up of migrants from subpopulation j each generation.

The first sentence suggests a backwards-time convention, but the last sentence could be interpreted as forwards-time wording. So is ms really is using forwards time conventions for the indices of `M` (and `m`)?

The 2007 fix makes no sense otherwise. Further, if the backwards convention applied, the Gutenkunst OOA model would potentially run forever (or give an "infinite coalescent time" error) when migrants entered an already merged population and could not escape (I tried running 100 reps with 300 chroms and it completed just fine). Finally, I dug further into the code. See below from `msdir/streec.c`, where `migm` is the migration matrix (and I guess the diagonal elements are temp variables, perhaps column or row sums, although I didn't figure that bit out yet).

```c
	    } else if (event == 'm') {	/* migration event */
		x = mig * ran1();
		sum = 0.0;
		for (i = 0; i < nchrom; i++) {
		    sum += migm[chrom[i].pop][chrom[i].pop];
		    if (x < sum)
			break;
		}
		migrant = i;
		x = ran1() * migm[chrom[i].pop][chrom[i].pop];
		sum = 0.0;
		for (i = 0; i < npop; i++) {
		    if (i != chrom[migrant].pop) {
			sum += migm[chrom[migrant].pop][i];
			if (x < sum)
			    break;
		    }
		}
		source_pop = i;
		config[chrom[migrant].pop] -= 1;
		config[source_pop] += 1;
		chrom[migrant].pop = source_pop;
	    }
```

Help?

molpopgen: This is a bit tricky to wade through.  Let me know if you don't figure it out--I may be able to dust off my memories here.

grahamgower: Thanks @molpopgen. I've woken up this morning with the idea that I should treat this as a black box and just construct a simple model demographic model to test it: if I guess the migration convention correctly, samples will coalesce, and if I guess incorrectly there will be an inifinite waiting time.

grahamgower: ms is using the forwards-time semantics/language for all migration indices

grahamgower:
I'm going to squash and merge this. Its too big to expect an in-depth code review and I'm fairly confident it's doing the right thing (>1000 lines of tests).

PR #105, opened 2020-11-18
convert deme graph to python code using demes API

grahamgower:
If you have a yaml file (or any demography convertable into a deme graph, like a stdpopsim demography) and you want to parameterise that model via the python API, you'd need to rewrite it using the python API first. The script here writes this python code automatically. I'm not sure if we should include something like this (or the exact form/location), but it seems useful.

```sh
$ python demes/convert/demes_.py examples/gutenkunst_ooa.yml > /tmp/gutenkunst_ooa.py
$ black /tmp/gutenkunst_ooa.py
reformatted /tmp/gutenkunst_ooa.py
All done! ✨ 🍰 ✨
1 file reformatted.
$ cat /tmp/gutenkunst_ooa.py
```
```python
g = demes.DemeGraph(
    description=(
        "The Gutenkunst et al. (2009) three-population model of human history. "
        "Time is given in years in the past. Choices to make, 1) which ends of "
        "the epochs should start and end times describe, 2) once that's decided, "
        "initial and final sizes should correspond to start and end time, resp. "
        "Currently, I'm looking backward in time, so start time is more recent "
        "and end time is the epoch interval end point farther in the past."
    ),
    time_units="years",
    generation_time=25,
    doi="10.1371/journal.pgen.1000695",
)
g.deme(
    id="ancestral",
    description="Equilibrium/root population",
    end_time=220000.0,
    initial_size=7300,
)
g.deme(
    id="AMH",
    description="Anatomically modern humans",
    ancestors=["ancestral"],
    end_time=140000.0,
    initial_size=12300,
)
g.deme(
    id="OOA",
    description="Bottleneck out-of-Africa population",
    ancestors=["AMH"],
    end_time=21200.0,
    initial_size=2100,
)
g.deme(
    id="YRI",
    description="Yoruba in Ibadan, Nigeria",
    ancestors=["AMH"],
    initial_size=12300,
)
g.deme(
    id="CEU",
    description="Utah Residents (CEPH) with Northern and Western European Ancestry",
    ancestors=["OOA"],
    initial_size=1000,
    final_size=29725,
)
g.deme(
    id="CHB",
    description="Han Chinese in Beijing, China",
    ancestors=["OOA"],
    initial_size=510,
    final_size=54090,
)
g.symmetric_migration(demes=["YRI", "OOA"], rate=0.00025)
g.symmetric_migration(demes=["YRI", "CEU"], rate=3e-05)
g.symmetric_migration(demes=["YRI", "CHB"], rate=1.9e-05)
g.symmetric_migration(demes=["CEU", "CHB"], rate=9.6e-05)
```

### DIFF
--- a/demes/convert/demes_.py
+++ b/demes/convert/demes_.py
@@ -1,0 +1,155 @@
+import textwrap
+
+import demes
+
+
+def to_demes(deme_graph: demes.DemeGraph):
+    def str_fix(string, indent=0, width=80):
+        """
+        Add quotes around string, and wrap long strings.
+        """
+        if not isinstance(string, str):
+            return string
+        if len(string) > width - indent:
+            indent2 = indent + 4
+            lines = textwrap.wrap(string, width=width - indent2)
+            string = (' "\n' + indent2 * " " + '"').join(lines)
+            string = "(\n" + indent2 * " " + f'"{string}"\n' + indent * " " + ")"
+        else:
+            string = f'"{string}"'
+        return string
+
+    d = deme_graph.asdict_compact()
+    py_str = []
+    py_str.append("g = demes.DemeGraph(")
+    for key in [
+        "description",
+        "time_units",
+        "generation_time",
+        "doi",
+        "selfing_rate",
+        "cloning_rate",
+    ]:
+        val = d.get(key)
+        if val is not None:
+            val = str_fix(val, indent=4)
+            py_str.append(f"    {key}={val},")
+    py_str[-1] = py_str[-1][:-1]  # remove trailing comma
+    py_str.append(")")
+
+    for deme_id, deme_attr in d["demes"].items():
+        py_str.append("g.deme(")
+        py_str.append(f'    id="{deme_id}",')
+        for key in [
+            "description",
+            "ancestors",
+            "proportions",
+            "start_time",
+            "end_time",
+            "initial_size",
+            "final_size",
+            "selfing_rate",
+            "cloning_rate",
+        ]:
+            val = deme_attr.get(key)
+            if val is not None:
+                val = str_fix(val, indent=4)
+                py_str.append(f"    {key}={val},")
+
+        epochs = deme_attr.get("epochs")
+        if epochs is not None:
+            py_str.append("    epochs=[")
+            for epoch in epochs:
+                py_str.append(8 * " " + "Epoch(")
+                for key in [
+                    "start_time",
+                    "end_time",
+                    "initial_size",
+                    "final_size",
+                    "size_function",
+                    "selfing_rate",
+                    "cloning_rate",
+                ]:
+                    val = epoch.get(key)
+                    if val is not None:
+                        if isinstance(val, str):
+                            val = f'"{val}"'
+                        py_str.append(12 * " " + f"{key}={val},")
+                py_str[-1] = py_str[-1][:-1]  # remove trailing comma
+                py_str.append("        ),")
+            py_str[-1] = py_str[-1][:-1]  # remove trailing comma
+            py_str.append("    ],")
+
+        py_str[-1] = py_str[-1][:-1]  # remove trailing comma
+        py_str.append(")")
+
+    migrations = d.get("migrations")
+    if migrations is not None:
+        symmetric = migrations.get("symmetric", [])
+        for migration in symmetric:
+            py_str.append("g.symmetric_migration(")
+            for key in [
+                "demes",
+                "start_time",
+                "end_time",
+                "rate",
+            ]:
+                val = migration.get(key)
+                if val is not None:
+                    if isinstance(val, str):
+                        val = f'"{val}"'
+                    py_str.append(f"    {key}={val},")
+            py_str[-1] = py_str[-1][:-1]  # remove trailing comma
+            py_str.append(")")
+
+        asymmetric = migrations.get("asymmetric", [])
+        for migration in asymmetric:
+            py_str.append("g.migration(")
+            for key in [
+                "source",
+                "dest",
+                "start_time",
+                "end_time",
+                "rate",
+            ]:
+                val = migration.get(key)
+                if val is not None:
+                    if isinstance(val, str):
+                        val = f'"{val}"'
+                    py_str.append(f"    {key}={val},")
+            py_str[-1] = py_str[-1][:-1]  # remove trailing comma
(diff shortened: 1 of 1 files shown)


jeromekelleher:
What about doing this as JSON instead? This has come up before, but I think having an explicit, low-level dictionary/JSON encoding of the DemeGraph would be useful for a lot of reasons.

grahamgower:
I think you might be pleasantly surprised by the `DemeGraph.asdict()` and `DemeGraph.asdict_compat()` methods. :) The latter is the exact data that is written to a yaml by the `dump()` function, but can instead be passed to the standard lib's `json.dump()`.

jeromekelleher:
Aha! Well, do we really need to output Python code then?

grahamgower:
Give me a few more minutes, and I'll show you where this is going... (and then you can suggest a better approach!)

grahamgower:
Ok, so suppose you want to reinfer some of the parameters you can see in a fixed-parameter yaml file. As a concrete example, suppose you've constructed a new statistic that is sensitive to archaic admixture, and you want to test how well it works to reinfer the proportion(s) of archaic introgression in the Jacobs et al. model with an ABC. Well, one can just convert the stdpopsim implementation to demes API code (instead of rewriting it manually!), and make it a function that accepts the desired parameters in order to output new yaml files drawn from the parameter distributions. The first stage is now fairly trivial (https://gist.github.com/grahamgower/8fd8f1c3f59194b6df2e86666230d076). The code in that gist outputs 100,000 replicate yaml files. In many cases we probably don't even need the yaml files, but let's assume we're going to feed them to a simulator which has no python API.

It takes about 40 mins to construct and write these to my SSD.
```
$ time python /tmp/jacobs_papuans.py
....................................................................................................
real    39m10,909s
user    39m1,808s
sys     0m2,481s
```

It looks like each file uses up a 4kb block on my ext4 filesystem, so this takes exactly 400 Mb of space.
```
$ ls /tmp/jacobs_pulse_models/|wc -l
100000
$ du -hs /tmp/jacobs_pulse_models/
391M    /tmp/jacobs_pulse_models/
$ du -ks /tmp/jacobs_pulse_models/
400000  /tmp/jacobs_pulse_models/
```

And compresses down to 10 Mb.
```
$ cd /tmp/
$ time tar czf jacobs_pulse_models.tar.gz jacobs_pulse_models/

real    0m2,211s
user    0m2,283s
sys     0m0,557s
$ ls -la jacobs_pulse_models.tar.gz
-rw-r--r-- 1 grg grg 10430070 Nov 18 13:02 jacobs_pulse_models.tar.gz
$ ls -lh jacobs_pulse_models.tar.gz
-rw-r--r-- 1 grg grg 10M Nov 18 13:02 jacobs_pulse_models.tar.gz
```

grahamgower:
Oh, and I just timed the same thing without writing any files and it took only 16 seconds.

jeromekelleher:
> Well, one can just convert the stdpopsim implementation to demes API code (instead of rewriting it manually!), and make it a function that accepts the desired parameters in order to output new yaml files drawn from the parameter distributions.

I see what you're doing now, and it's definitely an important application. What I don't get, though, is why you want to have a whole-new hard-coded version of the model in your Python function rather than modifying a copy of the original deme graph. So, something like
```
base_graph = stdpopsim.HomSap.JacobsPapuan_Whatever() # however we get a DemeGraph of a stdpopsim model, I forget
for alpha in make_alphas():
    g = base_graph.copy() 
    # Now overwrite the values in base_graph for the new values we've generated. We wouldn't actually do it 
    # like this, there'd need to be some better way of accessing the invidivual events that doesn't require us 
    # to hard code the times.
    g.pulse(source="Nea1", dest="Ghost", time=1853.0, proportion=alpha[0])
    g.pulse(source="Den2", dest="Papuan", time=1575.8620689655172, proportion=alpha[1])
    g.pulse(source="Nea1", dest="CHB", time=1566.0, proportion=alpha[2])
    g.pulse(source="Nea1", dest="Papuan", time=1412.0, proportion=alpha[3])
    g.pulse(source="Den1", dest="Papuan", time=1027.5862068965516, proportion=alpha[4])
    g.pulse(source="Nea1", dest="CHB", time=883.0, proportion=alpha[5])
```

See what I mean? Outputting code that people then modify is quite dangerous as it's easy to introduce mistakes (say I accidentally add a zero to some parameter in the file without realising). I think it would be much better to have a good API for making small modifications to an existing Graph.

jeromekelleher:
The 40 minutes is bananas! Where's the time being spent? It shouldn't take anything remotely like that long to write a few thousand 4K files.

grahamgower:
Perhaps this particular example lends itself to directly modifying parameters in the graph. But I worry that it would be easy to change the graph directly and leave it in an invalid state, particularly when there are complicated interactions between parameters (such as one deme split time being bounded by another split time: one split time in a yaml file could correspond with many dependent variables in the deme graph). To avoid this, the simplest thing to do is rebuild the graph from scratch.

Maybe it would be reasonable to rebuild from the dict/json representation instead, as you say. Right now this would be problematic, because a small implementation detail in our asdict_compact() method would change the parameters available for tuning. We guarantee `load()`/`dump()` produce "equivalent" models, but in practice the implementations of the model will often differ because we remove redundant information during the `asdict_compact()` call part of the `dump()` procedure. However, implementation details matter for making a parameterised model. This dirty process of spitting out python code provides a concrete implementation. But it's not pretty, I agree.

Probably the best way forward here, would be to modify parameters in the raw dict we get from strictyaml when it loads a file (before graph construction). So the process here is that one needs to write out a yaml file first, in order to obtain an specific model implementation, and then the names of the modifyable parameters will be properly defined. I'm ok with this, but it's also kinda clunky.

> The 40 minutes is bananas! Where's the time being spent? It shouldn't take anything remotely like that long to write a few thousand 4K files.

I guess the bowels of strictyaml (and it's dependency, ruamel.yaml) are very slow? Calling asdict_compact() in addition to building the graphs is only 30 seconds.

jeromekelleher:
That all makes sense @grahamgower, I think we should probably explore these options before resorting to the code-generation sledgehammer.

> I guess the bowels of strictyaml (and it's dependency, ruamel.yaml) are very slow? 

Must be, that's horrendously slow for just writing out some text files. We'd want to profile this at some point and see if we can cut some of the processing out.

Issue #106, opened 2020-11-18
API for epochs: remove Deme.add_epoch()?


grahamgower:
Defining epochs in the python API is a little bit inconsistent, and potentially confusing.

 * Firstly, we have the following construction methods on a DemeGraph: `deme()`, `pulse()`, `migration()`, but then we have `Deme.add_epoch()`. So the naming is inconsistent. We *could* rename it to `epoch()`, so one would do `g.deme("A", ...); g["A"].epoch(...)`, but...
 * The `add_epoch()` method currently requires the deme to already have at least one epoch defined. Naturally, an epoch is inserted when a deme is created with `deme()`. But by default this epoch spans the life time of the deme, so no new epochs can be added anyway.
 * So actually, it tends to not be useful/necessary to call the `add_epoch()` method at all. `DemeGraph.deme()` in fact accepts an `epochs` parameter which is a list of `Epoch`, and while this is also inconsistent with the other API construction methods, it provides much needed flexibility and convenience (e.g. see usage in `demes.loads()`).

So, is it possible to fix `add_epoch()` to make it useful and consistent (and perhaps remove the `epochs` arg to the `deme()` method), or should we just remove `add_epoch()`? I think making `add_epoch()` useful would be hard work though, because I think that `deme()` inserting a single epoch spanning the whole lifetime is very sensible (because maybe no more epochs get added), and so `add_epoch()` would thus need to accept non-temporally-ordered epochs which may also overlap previously added epochs.

apragsdale:
This might be quite similar to #97, where it's a nice idea for dynamically or incrementally building a demography, but ends up confusing users and cluttering the code. We could just remove it or make it not visible in the API, and require users to define every `epoch` of a deme in one go when defining the deme.

I think it was a useful to see how that incremental type of demography construction could go, but in the end it's perhaps unnecessary and ends up cluttering the code with multiple ways of defining deme epochs.

apragsdale:
Closed by #110

PR #110, opened 2020-11-18
Remove incremental building of DemeGraph

### LINKED ISSUE #97: DemeGraph methods: split(), branch(), admix(), and merge()

apragsdale:
Closes #97, #106 

work in progress, so far only removed `add_epoch( )`


apragsdale:
This PR completely removes the "incremental build" of a demography. It simplifies construction of a demography down quite a bit, so that all epochs and relationships between demes are set when we first call `g.deme(...)`.

It doesn't prohibit us from adding methods to incrementally define demographies in the future, but this gets us closer to the cleanest and simplest set of functions to define a demography using the API.

demes/demes.py
```diff
@@ -894,25 +838,41 @@ def deme(
...
-            if epochs[0].size_function is None:
-                if epochs[0].initial_size == epochs[0].final_size:
-                    epochs[0].size_function = "constant"
+            # fill in all attributes of epochs
+            for i in range(len(epochs)):
+                # assign default values to epoch attributes if needed
+                if epochs[i].selfing_rate is None:
+                    epochs[i].selfing_rate = selfing_rate
+                if epochs[i].cloning_rate is None:
+                    epochs[i].cloning_rate = cloning_rate
```
grahamgower, line 847: Is this bit needed here? Do these get set in the `Deme.__attrs_post_init__`?

demes/demes.py
```diff
@@ -519,12 +510,20 @@ def __attrs_post_init__(self):
                 raise ValueError("ancestors and proportions must have same length")
             if self.id in self.ancestors:
                 raise ValueError(f"{self.id} cannot be its own ancestor")
+        # every deme must have at least one epoch
```
apragsdale, line 513: Should these be a @epochs.validator decorated function, since we're checking properties of the `epochs` list we passed to `Deme`?

grahamgower (approved):
Beautiful! Deleting code is so much more rewarding than writing it!

demes/demes.py
apragsdale: You're right - those are redundant and can be removed.

demes/demes.py
grahamgower: I don't think it makes any difference either way. So it's up to you as to whether you think it's cleaner to have one validation function, or more than one.

apragsdale: Gotcha. Ok, going to leave as here and merge when checks pass

PR #120, opened 2020-11-19
add load_asdict() to load YAML/JSON as dict

### LINKED ISSUE #63: think about parameterised models
Once the dust settles, we'll want to add more complexity. :P

Currently, demes deals with **concrete** demographic models, ie. those with fixed parameter values. It would be really neat if we could also support **parameterised** demographic models, where parameters have a distribution and optionally depend on other parameters (eg. for an upper or lower bound). The motivation here, is to draw values from the distribution for simulation, do parameter inference using the parameterised model as a prior, or even output an inferred posterior parameterised demographic model.

This would need at least: (1) additional syntax in the yaml file, (2) a dependency graph for the parameters, and (3) an internal representation of parameterised models, from which it is cheap to construct many concrete models. We should consider carefully the design of the current version of ``demes``, to avoid unnecessarily making this harder in the future.

grahamgower:
load_asdict() allows a user to programmatically modify the model before
it gets built into a graph with demes.Graph.fromdict(). This facilitates
easy construction of parameterised models.

Closes #63.

### DIFF
--- a/demes/__init__.py
+++ b/demes/__init__.py
@@ -9,4 +9,4 @@ except ImportError:
     pass
 
 from .demes import Epoch, Migration, Pulse, Deme, Graph, Split, Branch, Merge, Admix
-from .load_dump import load, loads, dump, dumps
+from .load_dump import load_asdict, loads_asdict, load, loads, dump, dumps
--- a/demes/load_dump.py
+++ b/demes/load_dump.py
@@ -9,6 +9,47 @@ import demes
 from .schema import deme_graph_schema
 
 
+def loads_asdict(string, *, format="yaml"):
+    """
+    Load a YAML or JSON string into a dictionary of nested objects.
+    The keywords and structure of the string are defined by the
+    :ref:`schema <sec_schema>`.
+
+    :param str string: The string to be loaded.
+    :param str format: The format of the input string. Either "yaml" or "json".
+    :return: A dictionary of nested objects, with the same data model as the
+        YAML or JSON input string.
+    :rtype: dict
+    """
+    if format == "json":
+        d = json.loads(string)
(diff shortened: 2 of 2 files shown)


grahamgower:
Suppose we wanted to reinfer the time parameters of the Gutenkunst et al. OOA model using an ABC. The script below samples from the priors, and outputs 100,000 non-compact JSON demes graphs. The assumption is, these JSON files will be passed to a simulator that can't be used with a python API. The script took 40 seconds to run and a compressed tar of the resulting JSON files is 19 Mb.
```python
import pathlib

import numpy as np
import demes

n_reps = 100_000
output_path = pathlib.Path("/tmp/gutenkunst_time_models")
output_path.mkdir(parents=True, exist_ok=True)
d = demes.load_asdict("examples/gutenkunst_ooa.yml")
rng = np.random.default_rng()

# OOA splits into CEU/CHB
T0 = rng.uniform(15, 50, size=n_reps)
# AMH splits into OOA/YRI
T1 = rng.uniform(T0 + 10, T0 + 150, size=n_reps)
# ancestral size change
T2 = rng.uniform(T1 + 50, T1 + 150, size=n_reps)

for j, (t0, t1, t2) in enumerate(zip(T0, T1, T2)):
    if j % 1000 == 0:
            print(".", end="", flush=True)

    # update model parameters
    d["demes"]["OOA"]["end_time"] = t0
    d["demes"]["AMH"]["end_time"] = t1
    d["demes"]["ancestral"]["end_time"] = t2

    g = demes.Graph.fromdict(d)
    demes.dump(g, output_path / f"{j}.json", format="json", compact=False)
```

apragsdale (approved):
Looks great!

Issue #131, opened 2020-11-23
strictyaml permits comma-separated fields, but these are not YAML


grahamgower:
E.g.
```yaml
demes:
- id: A
  initial_size: 1000
- id: B
  initial_size: 1000
  ancestors: A
```
Is more correctly expressed as:
```yaml
demes:
- id: A
  initial_size: 1000
- id: B
  initial_size: 1000
  ancestors:
  - A
```
I guess this is fine if we're happy to convert comma-separated fields into lists ourselves. But any other consumers of the YAML files will have to do likewise.

jeromekelleher:
I would argue for sticking with the yaml spec, such as it is. Ironic that strictyaml decided to add its own extensions...

apragsdale:
This was closed by #132, yes?

grahamgower:
#132 didn't change the behaviour. We still use comma-separated fields in the YAML.

Issue #122, opened 2020-11-20
python yaml libraries: strictyaml, PyYAML, ruamel, ruyaml, ...


grahamgower:
The Python/`YAML` ecosystem is fragmented and confusing, so I've opened this issue to help clear things up. Currently, we're using the `strictyaml` library to validate the input YAML via a schema. `strictyaml` uses the `ruamel` library to do the loading/dumping. `strictyaml`'s documentation is an opinion piece rather than real documentation, and if we're going to do schema validation for both YAML and JSON input, the validation is better done at an intermediate layer rather than separately for JSON and YAML. So probably we can/should just load the YAML ourselves and drop `strictyaml`.

* [PyYAML](https://github.com/yaml/pyyaml) has been around for many years and offers `load()`/`dump()` functionality. There's a pure python implementation, and optional bindings to the C-based libyaml. There are minor differences between the C-based and Python-based loaders/dumpers. The C-based loader is not used transparently when available (it has to be used explicitly the python code). `PyYAML` only supports `YAML v1.1`. Supports python 2 and 3.
* [ruamel](https://yaml.readthedocs.io/en/latest/) is a fork of `PyYAML` that was created in response to a lack of activity in `PyYAML`, and the desire to fix many open issues, including the addition of support for `YAML v1.2`. The v1.2 spec appears to fix some odd corner cases in v1.1. `ruamel` has also introduced a new API, and is deprecating API compatibility with `PyYAML`. I couldn't find any API reference docs for the new API (just some discussion and a few examples). It's hard to tell for sure, but it seems that some of the desirable fixes are only available with the python-based loader/dumper. Supports python 2 and 3.
* [ruyaml](https://github.com/pycontribs/ruyaml) is a recent fork of `ruamel` that was created [in response to a lack of openness in `ruamel`](https://github.com/pycontribs/ruyaml/issues/1). It's still very new, so it's not clear how this will develop. The API is the same as for `ruamel` (both PyYAML and new API available), but they've already restructured the sources and dropped python < 3.6 support. Support for the C-based libyaml loader/dumper has been ([temporarily?](https://github.com/pycontribs/ruyaml/issues/6)) removed.

[...]

PR #132, opened 2020-11-23
replace strictyaml with ruamel.yaml

grahamgower:
WIP: still need to remove `demes/schema.py` and fix the docs that reference it.

See #122 and #131.


grahamgower:
Not sure what the deal is with newcomer `ruyaml`, but I could seem to import it after installing the package. So I've instead removed `strictyaml` in favour of directly using `ruamel.yaml`.

apragsdale (approved):
Looks good!

fopina:
> Not sure what the deal is with newcomer `ruyaml`, but I couldn't seem to import it after installing the package. So I've instead removed `strictyaml` in favour of directly using `ruamel.yaml`.

I'm also follow that package and the version <0.20 wasn't yet renamed (so package was still the old ruamel.yml.
It is now renamed AFAIK, would be good to see it being adopted :)  (disclaimer: not part of the project, but old user of ruamel.yml)

PR #135, opened 2020-11-24
Add hypothesis strategy for generating a Graph.

### LINKED ISSUE #24: generate random input graphs, for testing
The structure for the yaml file is reasonably settled now, so it shouldn't be too hard to generate random files.

grahamgower:
Closes #24.

---
I spent today learning about the [`hypothesis`](https://pypi.org/project/hypothesis/) library. This is just a proof of concept so far, but seems to work rather well. Probably we could use this in a lot more tests, but it might need some tweaks to avoid blowing out the test times.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -64,10 +64,10 @@ def isclose(
 
 
 def isclose_deme_proportions(
-    a_ids: Optional[List[ID]],
-    a_proportions: Optional[List[Proportion]],
-    b_ids: Optional[List[ID]],
-    b_proportions: Optional[List[Proportion]],
+    a_ids: List[ID],
+    a_proportions: List[Proportion],
+    b_ids: List[ID],
+    b_proportions: List[Proportion],
     *,
     rel_tol=_ISCLOSE_REL_TOL,
     abs_tol=_ISCLOSE_ABS_TOL,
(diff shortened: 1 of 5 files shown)


grahamgower:
It occurs to me that we might want to export this (or something like it) under a `demes.testing` namespace, so that `demes` library users can benefit from this too.

apragsdale:
That looks really useful - has it found any corner cases that raise errors (or failed to) that we've missed?

grahamgower:
Not much. Just a couple of things that are also fixed as part of this PR (e.g. we shouldn't allow demes with `id=""`).

apragsdale:
Right, I saw some of those extra tests - they make sense and are good to include here. Any idea why the one test on ubuntu, 3.6 is failing? I couldn't trace what was the matter there

Nice work!

grahamgower:
> Any idea why the one test on ubuntu, 3.6 is failing? I couldn't trace what was the matter there - just complaining about data generation being quite slow?

Tests that use the `@given(...)` decorator will, by default, run the test with 100 different inputs. So, yeah, its just complaining that this is slow. I need to do more reading to see if how I've constructed the graphs() "strategy" is harming performance, or if there's something else we can do to avoid the failing health check. I'm not sure if the tests are deterministic either---we'll probably want to nail that down too.

jeromekelleher (approved):
Awesome! I love this, it'll be super useful.

grahamgower:
I've marked it as a draft for now because I still think this needs more work, particularly in regard to making it play nicely with CI. It will also conflict with #138, and I'd prefer to prioritise merging that PR over this one.

grahamgower:
This was such a nightmare to rebase, I've wound back a couple of things and opened a new PR in #148.

Issue #111, opened 2020-11-19
Graph.deme should return the newly created Deme


jeromekelleher:
It's often handy to have a reference to the thing you've created, and it does no harm when you don't use it.

grahamgower:
I'm not opposed to this, but the deme is available using `DemeGraph.__getitem__()`.
```
>>> import demes
>>> g = demes.DemeGraph(description="test", time_units="generations")
>>> g.deme("A", initial_size=10)
>>> g["A"]
Deme(id='A', description=None, ancestors=None, proportions=None, epochs=[Epoch(start_time=inf, end_time=0, initial_size=10, final_size=10, size_function='constant', selfing_rate=None, cloning_rate=None)], selfing_rate=None, cloning_rate=None)
```
We also have a `DemeGraph.__contains__()` that facilitates checking if a deme is in the graph.
```
>>> "A" in g
True
```

If we *did* return the newly created `Deme`, should we also have `migration()` return the `Migration`, and `pulse()` return the `Pulse`?

jeromekelleher:
> If we did return the newly created Deme, should we also have migration() return the Migration, and pulse() return the Pulse?

I think it's a good idea for all these methods to return the newly created item. It's a common pattern to want to do something with the thing you've created, and you shouldn't have to type your string ID twice to do this. 

It's not like we want the return value for anything else anway, is it?

grahamgower:
Ok, makes sense. It's an easy change and doesn't compilcate anything, so I'm happy for it to go in.

noscode:
Should `symmetric_migration()` return the list of created migrations?

grahamgower:
Yeah, that seems reasonable.

apragsdale:
Closed by #137.

PR #137, opened 2020-11-25
Return newly created objects in Graph methods

noscode:
Solves #111.

I have add the return of newly created objects in :

- `Graph.deme` returns `Deme` object,
- `Graph.migration` returns `Migration` object,
- `Graph.pulse` returns `Pulse` object,
- `Graph.symmetric_migration` returns list of `Migration` objects.

Also I have put simple type checks in tests: `TestGraph.test_newly_created_objects_return` function.

[UPDATE] I have also changed type of Graph._deme_map from `Mapping` to `Dict` as there was an error from mypy that Mapping is not mutable and it is updated in `Graph.deme` method. I have found that the [best solution](https://stackoverflow.com/questions/60456140/unsupported-target-for-indexed-assignment-with-mypy-depending-on-type-hinting) is TypedDict but it is available since python3.8 only.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1,5 +1,5 @@
 import typing
-from typing import List, Mapping, Union, Optional
+from typing import List, Union, Optional, Dict
 import itertools
 import math
 import copy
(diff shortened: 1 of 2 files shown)


demes/demes.py
```diff
@@ -844,6 +846,7 @@ def deme(
         )
         self._deme_map[deme.id] = deme
         self.demes.append(deme)
+        return self._deme_map[deme.id]
```
grahamgower, line 849: How about just returning `deme` here?

grahamgower (approved):
This is fantastic, thanks @noscode! I only have one tiny suggestion below...

demes/demes.py
noscode: Sure, missed that one..

apragsdale:
Looks great! Thank you @noscode!

Issue #58, opened 2020-10-27
describe best-practices for specifying a demes graph


grahamgower:
E.g.

- for two extant populations, define (Ancestral, (child1, child2)), rather than (child1, (child1 continuation, child2).
- #46

grahamgower:
Demes should be defined from the root(s) of the graph downwards. This is because when specifying a deme's ancestors, the ancestor demes must already have been defined.

grahamgower:
I think it's clearer when yaml files use `start_time` rather than `end_time`. Sometimes using `start_time` more will result in a less compact definition, but I think that's fine.
E.g. in the docs we have (`docs/tutorial_examples/two_epoch.yaml`):
```
description: Single-population two-epoch demography.
time_units: generations
demes:
  Deme_name:
    description: A deme that doubles in size 100 generations ago.
    epochs:
    - initial_size: 1000
      end_time: 100
    - initial_size: 2000
      end_time: 0
```
Which I think would be clearer as (`/tmp/a.yml`):
```
description: Single-population two-epoch demography.
time_units: generations
demes:
  Deme_name:
    description: A deme that doubles in size 100 generations ago.
    initial_size: 1000
    epochs:
    - start_time: 100
      end_time: 0
      initial_size: 2000
```

It doesn't help that when we `dump()`, the output prefers `end_time`.
```
>>> import demes
>>> a = demes.load("/tmp/a.yml")
>>> b = demes.load("docs/tutorial_examples/two_epoch.yaml")
>>> a.isclose(b)
True
>>> print(demes.dumps(a))
description: Single-population two-epoch demography.
time_units: generations
demes:
  Deme_name:
    epochs:
    - end_time: 100
      initial_size: 1000
    - end_time: 0
      initial_size: 2000
    description: A deme that doubles in size 100 generations ago.
```

grahamgower:
@apragsdale makes the very good point that "it makes more sense to specify end times in general and have derived demes inherit their start time from their ancestors, instead of having ancestors inherit end times from derived demes' start times. There's some notion of the flow of time that makes more sense in the first case, I think."

apragsdale:
[...]

PR #138, opened 2020-11-25
Add asdict_simplified

apragsdale:
This reintroduces `asdict_simplified()` using Graham's start at reformatting it. It also tries to address some of the issues from #58 - flexible specification, but "canonical" simplified output.

Would love suggestions here

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -120,7 +120,9 @@ class Epoch:
     """
 
     start_time: Optional[Time] = attr.ib(default=None, validator=optional(non_negative))
-    end_time: Time = attr.ib(default=None, validator=[non_negative, finite])
+    end_time: Optional[Time] = attr.ib(
+        default=None, validator=optional([non_negative, finite])
+    )
     initial_size: Optional[Size] = attr.ib(
         default=None, validator=optional([positive, finite])
     )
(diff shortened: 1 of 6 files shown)


grahamgower:
Nice. Thanks for picking this back up @apragsdale, I really like it! Looks like it just needs a couple more tests. And I'm curious, what does it do if you have symmetric migration between say, 3 or 4 demes?

apragsdale:
> Looks like it just needs a couple more tests. And I'm curious, what does it do if you have symmetric migration between say, 3 or 4 demes?

Yes, more tests needed.. I'll work on that.

Current implementation only reduces symmetric migrations to pairwise, but if there are more demes that are all pairwise symmetric at same rate, I could maybe add a loop using `itertools.combinations` to check all possible combinations that start/end times and rates are all equal, for varying sizes.. need to think about it a bit

grahamgower:
Maybe just check nothing crazy happens, and ignore the difficult case for
now?

apragsdale:
This ends up reintroducing `cloning_rate` and `selfing_rate` at the graph-level for inheriting attributes and passing them up and down to/from epochs. Not sure if this is something we want to do...

grahamgower:
> This ends up reintroducing `cloning_rate` and `selfing_rate` at the graph-level for inheriting attributes and passing them up and down to/from epochs. Not sure if this is something we want to do...

I would prefer if the class attributes matched the data model for a fully-qualified graph. And if the goal is to output the smallest file, then maybe copying what the user specified isn't the right thing anyway? In any event, it's not so important that we have the simplification process doing the best possible job for our first release. In my opinion, the epoch start_time/end_time simplification you have in this PR is really valuable, and that should be the current focus.

To digress about inherited attributes though... How about instead using a "greedy" simplification rule? So, first at the Deme level, take the most-frequent attribute value across the epochs as the default. Then do the same thing at the Graph level, taking the most-frequent value across the demes. It's not perfect, but it should be good enough in the most common cases.

jeromekelleher:
> To digress about inherited attributes though... How about instead using a "greedy" simplification rule? So, first at the Deme level, take the most-frequent attribute value across the epochs as the default. Then do the same thing at the Graph level, taking the most-frequent value across the demes. It's not perfect, but it should be good enough in the most common cases.

Probably the thing to do is to use a parsimony algorithm - we think about inherited attributes as state changes on the object model tree, and the simplified thing we output one of the optimal assignments. I've got some Python code lying around for doing Hartigan parsimony, which will do this and I'm sure it'll be easy enough to adapt.

grahamgower:
> I've got some Python code lying around for doing Hartigan parsimony, which will do this and I'm sure it'll be easy enough to adapt.

Great! Sounds like it really should be its own PR then.

jeromekelleher:
> Great! Sounds like it really should be its own PR then.

Agreed - my suggestion would be to pare this back to start-time/end-time simplification and merge. We can follow up separately on inherited attributes.

apragsdale:
> Agreed - my suggestion would be to pare this back to start-time/end-time simplification and merge. We can follow up separately on inherited attributes.

Ok perfect - I'll pare this back, leaving those attributes at the epoch level, and we can handle that separately.

apragsdale:
Additionally, if there are more than two demes that have compatible symmetric migration (say demes A, B, and C all have symmetric migration with the same rates and start and end times), this returns a compact/simplified graph that lists A<->B, A<->C, B<->C, instead of reducing all three into the same symmetric migration specification. I've got something that is close to reducing those symmetric rates, but it's a bit messier requiring , so I'll open it as its own PR once this one is settled, to see if we like it.

grahamgower (approved):
LGTM!

PR #141, opened 2020-11-26
Merge n-way symmetric migrations

apragsdale:
I don't know if I've made this overly complicated, but the algorithm here in the `simplify_migration_rates()` function in `asdict_simplified()` iteratively searches for sets of n-way symmetric migrations within a given migration rate attribute set (rate, start_time, end_time), in a way that hopefully catches all n-way symmetric migrations, starting with those that include larger sets of demes.

It needs some more testing, but there are a few simple test scenarios in `test_demes.py`.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1676,9 +1676,6 @@ class Graph:
         """
         Return a simplified dict representation of the graph.
 
-        This function removes redundancies in the graph. Specifically, we ...
-        continue docs
-
         :param custom_attributes: List of additional attributes to simplify, which
             are not ``selfing_rate`` or ``cloning_rate``.
         """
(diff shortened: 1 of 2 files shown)


demes/demes.py
```diff
@@ -1224,20 +1242,68 @@ def simplify_migration_rates(data):
...
-                    sym = dict(demes=[dest, source], **opposite)
-                    del sym["source"]
-                    del sym["dest"]
-                    symmetric.append(sym)
+
+                k = [migration["rate"]]
+                try:
+                    k.append(migration["start_time"])
+                except KeyError:
+                    k.append(None)
```
grahamgower, line 1250: Here, you can use the idiom `k.append(migration.get("start_time"))` instead.

grahamgower: Or indeed `k = tuple(migration.get(key) for key in ("rate", "start_time", "end_time"))`

grahamgower (approved):
It's a bit complicated, but much less so than I was expecting. Nice job!

demes/demes.py
apragsdale: Oh, nice! Didn't know I could do that. Thanks

PR #146, opened 2020-11-27
fix version string for CI docs build

grahamgower:


grahamgower:
Merging this anyway, because there's no way to test until it's merged.

PR #148, opened 2020-11-27
Add hypothesis strategy for generating a Graph.

### LINKED ISSUE #24: generate random input graphs, for testing

### LINKED ISSUE #136: restrict the allowable format for deme IDs
Currently we allow just about anything. But this is likely a bad idea. E.g. I can imagine calling my deme ".inf" from the python API, and then after writing it to the yaml, the `id` might be parsed back in as a float. More generally though, deme IDs are probably going to be copied into unexpected places (I bet someone will put them into a filename), so we should be conservative here. Maybe we should check against a regex ike `[-_a-zA-Z]([-_a-zA-Z0-9])*`? Any reason to permit non ascii, or other symbols?

grahamgower:
And fix a bunch of the fallout.

Closes #24.
Closes #136.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -42,6 +42,12 @@ def unit_interval(self, attribute, value):
         raise ValueError(f"must have 0 <= {attribute.name} <= 1")
 
 
+def nonempty_str(self, attribute, value):
+    attr.validators.instance_of(str)(self, attribute, value)
+    if len(value) == 0:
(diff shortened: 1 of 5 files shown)


jeromekelleher (approved):
Nice, these all look like valuable clarifiations.

apragsdale (commented):
Looks good to me!

PR #152, opened 2020-12-01
simplify API by removing loads()/dumps()

### LINKED ISSUE #131: strictyaml permits comma-separated fields, but these are not YAML

grahamgower:
It's not so common to need to load/dump from/to a string. But if you
really need to do it, e.g. we use this for tests, it's quite easy to
use a temporary file.
Moreover, we didn't have a way to load/dump from/to a file-like object
that has a write() method. This is far more useful, so we add that here.
This also provides another option for strings via io.StringIO().

Also closes #131 while in the neighbourhood.

### DIFF
--- a/demes/__init__.py
+++ b/demes/__init__.py
@@ -9,4 +9,4 @@ except ImportError:
     pass
 
 from .demes import Epoch, Migration, Pulse, Deme, Graph, Split, Branch, Merge, Admix
-from .load_dump import load_asdict, loads_asdict, load, loads, dump, dumps
+from .load_dump import load_asdict, load, dump
--- a/demes/convert/msprime_.py
+++ b/demes/convert/msprime_.py
@@ -312,7 +312,12 @@ def from_msprime(
                         dest=name[k],
                         start_time=ddb_epoch.end_time,
                         end_time=ddb_epoch.start_time,
-                        rate=msp_mm[j, k],
+                        # Explicitly convert to float, so we don't leak a
+                        # singleton numpy float type. This can be a problem
+                        # for annoying serialisers that want to look at the
+                        # type instead of happily outputting a number.
+                        # I'm looking at you, ruamel.yaml!
+                        rate=float(msp_mm[j, k]),
                     )
                     migrations[(j, k)].append(m)
                 else:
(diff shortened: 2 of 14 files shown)


grahamgower:
Windows-specific failures from hypothesis-generated nonsense. */me shakes fist*

jeromekelleher:
I'm not sure I agree with this - I always find having a dumps/loads really handy and being forced to makes a StringIO annoying and tedious. Is it really a maintenance burden?

grahamgower:
There is a maintenance burden associated with keeping both load/dump and loads/dumps, unless one is a wrapper for the other. The existing `demes.load()` and `demes.dump()` functions are wrappers around the string versions, `demes.loads()` and `demes.dumps()`. I wanted to change it so that file and stream operations are first class, instead of going via strings.

For json, this makes little difference, as the standard lib is fairly sane for both `json.load`/`json.dump` and `json.loads()`/`json.dumps()`. But the yaml library (libraries) are purely stream oriented. This means that we end up going through strings twice, e.g. something like: `demes.dump() -> string - > demes.dumps() -> yaml.dump_shim() -> StringIO -> yaml.dump() -> file`. That's quite horrible, especially since yaml performance is already bad.

If you want, we can keep `demes.loads()` and `demes.dumps()`, but I reckon these should be the wrapper functions, not the other way around. I admit, they're useful for debugging and writing tests. But user code will (hopefully) not need them.

grahamgower:
In any event, there are additional issues that arose while I was working on this PR. So I'm going to break this up and try to fix each thing in separate PRs.

jeromekelleher:
> If you want, we can keep demes.loads() and demes.dumps(), but I reckon these should be the wrapper functions, not the other way around. I admit, they're useful for debugging and writing tests. But user code will (hopefully) not need them.

Sure, very happy for them to be wrappers around the stream version. As you say, they're mostly useful for debugging etc, so we don't need to worry too much about performance and so on.

PR #160, opened 2020-12-02
change load()/dump() to accept streams

### LINKED ISSUE #153: allow load()/dump() to work with streams
They currently only work with files. And the loads()/dumps() variants work only wth strings.

### LINKED ISSUE #154: change load()/dump() to work directly with files/streams, not via strings
See https://github.com/popsim-consortium/demes-python/pull/152#issuecomment-737063992.

grahamgower:
Also make the loads()/dumps() functions wrap these, instead of
the other way around.

Closes #153.
Closes #154.

### DIFF
--- a/demes/load_dump.py
+++ b/demes/load_dump.py
@@ -1,12 +1,27 @@
 """
 Functions to load and dump graphs in YAML and JSON formats.
 """
+import contextlib
 import json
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM

PR #168, opened 2020-12-03
Closes #166

### LINKED ISSUE #166: Make sure we don't allow competing continuous migrations
Currently, don't think there is a catch to make sure we haven't specified continuous migration between the same source/dest demes that have overlapping time intervals.

E.g., this should result in an error:
```YAML
migrations:
- symmetric:
  - demes: A, B
    rate: 0.01
  - demes: A, B
    rate: 0.02
```

apragsdale:
Checks non-overlap of migration start/end times if they involve the same source/dest. Am I missing any cases here?

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1567,6 +1567,22 @@ class Graph:
             end_time = time_lo
         else:
             self._check_time_intersection(source, dest, end_time)
+        for existing_migration in self.migrations:
+            if source == existing_migration.source and dest == existing_migration.dest:
+                if (
+                    start_time >= existing_migration.start_time > end_time
+                    or start_time > existing_migration.end_time >= end_time
+                    or existing_migration.start_time
+                    >= start_time
+                    > existing_migration.end_time
+                    or existing_migration.start_time
+                    > end_time
+                    >= existing_migration.end_time
+                ):
+                    raise ValueError(
+                        "new migration overlaps exisiting migration "
+                        f"from {source} to {dest}"
+                    )
         migration = Migration(
             source=source,
             dest=dest,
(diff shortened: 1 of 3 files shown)


apragsdale:
Ah, yes this seems to be failing from the hypothesis draws. @grahamgower is that an easy change to writing those tests that you had in mind? Or requires some major rewriting..?

grahamgower:
At the moment, `graphs()` adds all demes in a loop, then draws all migrations/pulses in another loop. It could be changed to draw mutations/pulses in the deme-adding loop, just between the currently added deme and the existing demes. It would then be easier to limit migrations to be non-time-overlapping for each deme pair. E.g. see how times are drawn in the `epochs_lists()` function above it, to draw a set of unique times, sort them, and then use them as non-overlapping time intervals.

apragsdale:
Ok, trying something here to avoid adding overlapping migrations to a random graph in the tests...

apragsdale:
Hey @grahamgower ... not sure what this error here means in build-deploy-docs: "Error: Input required and not supplied: token". Do we need to update something in setting up github actions?

grahamgower:
Ah, sorry about the docs action failure. I disabled that for forks, because github suggested there's a security issue with it (Ben Jeffery told me he'll come up with a way of doing this differently). So for now, pull requests from forks will get permission errors from the docs github action. If it gets too annoying, I'll temporarily delete the action. Or you can push a branch to the `popsim-consortium/demes-python` repo, and open a pull request from there (which will have the right permissions). This all applies to the demes-spec repo too, btw.

grahamgower (approved):
LGTM @apragsdale.

jeromekelleher (approved):
LGTM also

apragsdale:
Is this ok to merge and just ignore that error for now, then?

jeromekelleher:
Fine by me

PR #171, opened 2020-12-04
Simplified dict retains epochs

apragsdale:
Previous behavior was that if a deme had a single epoch, start/end times and sizes were dropped down to the deme level. This retains all epochs instead.


grahamgower (approved):
Looks good!

PR #172, opened 2020-12-04
defaults and defining demes

apragsdale:
This takes a stab at a few of the things we've been discussing, including requiring that a non-empty list of `epochs` must be given for each `deme`, and that attributes like `selfing_rate` and `cloning_rate` are given in a dict of `defaults` at the `graph` and `deme` level.

Definitely a work in progress and will need some cleaning up or rethinking perhaps.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -961,6 +961,10 @@ class Deme:
     :ivar epochs: A list of epochs, which define the population size(s) of
         the deme. The deme must be created with all epochs listed.
     :vartype epochs: list of :class:`.Epoch`
+    :ivar defaults: A dictionary of default attributes, such as cloning and
+        selfing rates, that are inherited by epochs if those attributes are
+        not specified in the epoch.
+    :vartype defaults: dict
     """
 
     id: ID = attr.ib(validator=[attr.validators.instance_of(str), valid_deme_id])
@@ -989,6 +993,7 @@ class Deme:
             iterable_validator=attr.validators.instance_of(list),
         )
     )
+    defaults: dict = attr.ib()
 
     @ancestors.validator
     def _check_ancestors(self, _attribute, _value):
@@ -1155,6 +1160,8 @@ class Graph:
     :ivar pulses: A list of migration pulses for the demography.
         Use :meth:`pulse` to add a pulse.
     :vartype pulses: list of :class:`.Pulse`
+    :ivar defaults: A dictionary of default attributes for demes and epochs.
+    :vartype defaults: dict
     """
 
     description: str = attr.ib(
@@ -1177,6 +1184,7 @@ class Graph:
     demes: List[Deme] = attr.ib(factory=list, init=False)
     migrations: List[Migration] = attr.ib(factory=list, init=False)
     pulses: List[Pulse] = attr.ib(factory=list, init=False)
+    defaults: dict = attr.ib(default={})
 
     def __attrs_post_init__(self):
         self._deme_map: Dict[ID, Deme] = dict()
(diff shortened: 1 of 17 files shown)


demes/demes.py
```diff
@@ -1412,78 +1429,62 @@ def deme(
...
-            if epochs[-1].end_time is None:
-                if end_time is not None:
-                    epochs[-1].end_time = end_time
+            selfing_rate = None
+        if "cloning_rate" in defaults:
+            cloning_rate = defaults["cloning_rate"]
+        elif "cloning_rate" in self.defaults:
+            cloning_rate = self.defaults["cloning_rate"]
+        else:
+            cloning_rate = None
```
grahamgower, line 1444: How about:
```suggestion
        cloning_rate = defaults.get("cloning_rate", self.defaults.get("cloning_rate"))
```
And similarly for `selfing_rate` above.

demes/demes.py
```diff
@@ -1412,78 +1429,62 @@ def deme(
...
+            raise ValueError("deme and first epoch start times do not align")
+
+        # fill in all attributes of epochs
+        for i in range(len(epochs)):
+            # set the start and end times based on surrounding demes
+            if epochs[i].start_time is None:
+                epochs[i].start_time = epochs[i - 1].end_time
+            if epochs[i].end_time is None:
+                if epochs[i + 1].start_time is None:
+                    raise ValueError("ambiguity about epochs' start/end times")
```
grahamgower, line 1462: The exception here should mention the deme id, and maybe also the indexes `i`/`i+1`?

demes/demes.py
```diff
@@ -1412,78 +1429,62 @@ def deme(
...
+                epochs[i].end_time = epochs[i + 1].start_time
+            # for each subsequent epoch, fill in start size, final size,
+            # and size function as necessary based on last epoch
+            if epochs[i].initial_size is None:
+                epochs[i].initial_size = epochs[i - 1].final_size
```
grahamgower, line 1467: It's not clear to me what happens here in the case of only one epoch.

demes/demes.py
```diff
@@ -1326,8 +1334,7 @@ def deme(
         end_time=None,
         initial_size=None,
         final_size=None,
```
grahamgower, line 1336: It looks like you've removed the calls to this function that use the start_time, end_time, initial_size, and final_size params. Was it your intention to remove the args here too?

grahamgower (approved):
I didn't read through all the tests, but this looks great! Thanks @apragsdale!

apragsdale:
Thanks for the comments Graham! A bit of work to do on this still - will get to it in a day or two though

grahamgower:
Hey @apragsdale, do you mind if I pick up this PR and move it forward? I'd really like to make this match the reference implementation in the spec repo, so we can push out an alpha release of demes-python ASAP.

apragsdale:
Hey Graham- yes, go for it! I forget what sort of shape it’s in.. so I hope it’s not too bad.

On 14 Jan 2021, at 2:24, Graham Gower <notifications@github.com> wrote:

﻿

Hey @apragsdale<https://github.com/apragsdale>, do you mind if I pick up this PR and move it forward? I'd really like to make this match the reference implementation in the spec repo, so we can push out an alpha release of demes-python ASAP.

apragsdale:
Picking this back up - would be great to get a couple sets of eyes on this to make sure we're implementing it according to the specs that we've settled on.

jeromekelleher:
I'm not really familiar enough with the existing code to give an informed review here I'm afraid @apragsdale. My take would be to get a version of this merged soon, and then add some more tests that would compare the results of the local parser here with that of the reference implementation across a range of inputs. That would give us good confidence the two are equivalent.

I think the docs build break is unrelated, probably something to do with the issue on PR builds from a while back. (So, nothing to worry about)

apragsdale:
Ok sounds good - I'll wait to see if @grahamgower notices anything obvious that needs fixing up, and then we can go forward with writing those tests to compare to the demes-spec.

grahamgower (approved):
LGTM!

Issue #182, opened 2021-01-27
require deme.start_time and epoch.end_time in input files


grahamgower:
This change is to follow the reference implementation. See https://github.com/popsim-consortium/demes-spec/issues/32

apragsdale:
My understanding was that the deme does not require a start time, except in scenarios where it has multiple ancestors. Otherwise it inherits the start time from the ancestor's end time, or the start time is set to inf if it has no ancestors. That was what I thought the takeaway of that linked issue was, and how it's implemented in the spec now. Am I wrong about that?

grahamgower:
Yeah, you're right. I opened this issue before looking into the gory details. The essential changes we need to make are to (1) disallow `start_time` in an epoch stanza, (2) allow the omission of `end_time` in the last epoch if the end time is `0`.

I made a start on this in #188. But I won't get back to it for a couple of days because I'll be away from the keyboard (again). So, you're welcome to jump on it and resolve anything that doesn't seem to match the behaviour of the `reference_implementation`.

PR #188, opened 2021-01-29
Require deme start_time, not epoch start_time.

grahamgower:
Related to #182.


grahamgower:
Ok, I think this is good to go. It fixes the most invasive things needed to conform to the spec.

tests/test_demes.py
```diff
@@ -1,9 +1,10 @@
 import unittest
 import copy
 import pathlib
-import json
 
-import jsonschema
+# import json
```
jeromekelleher, line 5: delete?

jeromekelleher (approved):
LGTM

tests/test_demes.py
grahamgower: You're right, it's probably better to just delete this. I thought maybe we'd reuse it again soon for validating against the spec, but probably we'll do something different (via CI) to compare with the spec and/or spec repo examples.

PR #195, opened 2021-02-01
Fix tests workflow to use numpy version in requirements.

grahamgower:
Msprime depends on numpy too, and the recent release of numpy 1.20.0 has
meant that 'conda install msprime' installs the latest numpy instead of
the version we have in requirements.txt. So this changes the 'conda install'
line in the tests workflow to also install numpy (with the correct version).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,9 +44,21 @@ jobs:
 
       - name: install dependencies
         run: |
+          echo "## conda list"
+          conda list
           conda install --yes $(grep numpy requirements.txt) $(grep msprime requirements.txt)
+          echo "## conda list"
+          conda list
           pip install -r requirements.txt
+          echo "## conda list"
+          conda list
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Why not bump the numpy version in requirements.txt? msprime should be robust to any new version of numpy (unless they've changed the ABI)

grahamgower:
> Why not bump the numpy version in requirements.txt? msprime should be robust to any new version of numpy (unless they've changed the ABI)

Because numpy 1.20.0 drops support for python 3.6 :)  #192

jeromekelleher:
Ah. Might be easier to drop py36 then? Messing with conda is rarely productive...

grahamgower:
Yeah, we might have an easier time if we drop python 3.6. But I think we should still pin the numpy version that is installed with conda.

grahamgower:
Picked up in #197.

PR #197, opened 2021-02-10
Remove moments converter.

### LINKED ISSUE #196: recent tests workflow failures with moments on python 3.9
Not sure what's going on here, but there are failures evident in today's dependabot PRs. I thought they were due to having multiple numpy versions installed, so opened #195 to fix that. But there's something else going on which is not at all obvious.

I can reproduce the test failures locally, after installing moments via
```pip install git+https://bitbucket.org/simongravel/moments@moments-1.0.8#egg=moments```.
However, the failures don't occur after installing moments from a locally cloned repository with `python setup.py install --user`.

grahamgower:
Closes #196.

Looks like the moments tests were hitting some code paths not tested elsewhere, so we'll need to add some more tests for that.


grahamgower:
I've cherry-picked the fix in #195 over to here, because that's still needed to resolve the numpy version mismatch.

apragsdale (approved):
Thanks for tackling this Graham. Looks good to me.

apragsdale:
I'm not sure exactly what the issue with numpy still is, if it was caused by moments but moments has now been removed. Also not sure if the fix I pushed out in moments-1.0.8 didn't end up solving that numpy version issue. But in any case, removing the moments dependency simplifies things greatly.

grahamgower:
There was a numpy issue in the demes CI workflow, in addition to the problem with `moments` build dependency. :(

PR #202, opened 2021-02-15
Initial outline of simple models.

jeromekelleher:
We want easy ways to create standard models. This is a first pass for discussion. Use like

```python3
import demes

# 10 population island model, each of size 100
g1 = demes.models.island_model([100] * 10, migration_rate=0.1)
# 3 population stepping stone model, of size 10, 20 and 30
g2 = demes.models.stepping_stone([10, 20, 30], migration_rate=0.1)
```

Using the ``demes.models``  namespace here seems like a nice way to do this.

### DIFF
--- a/demes/__init__.py
+++ b/demes/__init__.py
@@ -10,3 +10,4 @@ except ImportError:
 
 from .demes import Epoch, Migration, Pulse, Deme, Graph, Split, Branch, Merge, Admix
 from .load_dump import load_asdict, loads_asdict, load, loads, dump, dumps
+import demes.models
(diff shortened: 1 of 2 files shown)


grahamgower:
That looks nice and clean, and I think the `demes.models` namespace is a good one. It probably does belong here in Demes, but in any event I'll link to the related [stdpopsim issue](https://github.com/popsim-consortium/stdpopsim/issues/479).

jeromekelleher:
Great, I'll add some tests. Generic models always seemed like bloat in stdpopsim to me.

Issue #208, opened 2021-02-17
GraphBuilder API?


jeromekelleher:
This is going to be unpopular, so apologies in advance. 

I'm not a huge fan of the current API for building graphs incrementally in Python. I find the tension between being able to specify a graph concisely without all the unnecessary attributes and the desire to have a fully featured in-memory model of a Graph that you can be sure is non-redundant, etc, confusing.

Currently, we have things like
```python
    graph = demes.Graph(description="Island model", time_units="generations")
    for j, start_size in enumerate(size):
        graph.deme(
            id=f"pop_{j}", epochs=[demes.Epoch(start_size=start_size, end_time=0)]
        )
    graph.symmetric_migration(
        demes=[deme.id for deme in graph.demes], rate=migration_rate
    )
```

What if we had, instead,
```python
    gb = demes.GraphBuilder(description="Island model", time_units="generations")
    for j, start_size in enumerate(size):
        deme = gb.add_deme(id=f"pop_{j}")
        deme.add_epoch(start_size=start_size, end_time=0)
    gb.add_symmetric_migration(
        demes=[deme.id for deme in gb.demes], rate=migration_rate
    )
    graph = gb.finalise() # Or gb.bake(), or gb.resolve_all_the_things() or whatever
```

So, then the graph is considered to be fully read-only, and is guaranteed to have all the necessary attributes set. If you want to update an existing graph, you call ``gb = graph.builder()`` or something.

I think that ultimately the graphs will be fundamentally more useful if they are immutable; making code that reasons with them will be much easier in the long run. It also makes a natural point at which to run the graph resolution process, which we need for parsing. I don't see this currently with the current API?

Thoughts?

grahamgower:
I considered a factory-style class (separate to the underlying Graph, Deme, Epoch, etc. objects) when first sketching things out, but I think I discarded the idea largely because adding epochs incrementally started getting awkward. In your example above, what would `GraphBuilder.add_deme()` return? A `Deme`, complete with `add_epoch()` method, or a `DemeBuilder` class? It would probably be simpler if add_epoch() was instead a method of GraphBuilder (e.g. `def GraphBuilder.add_epoch(*, deme, start_size, end_time, ...):`.

[...]

Issue #201, opened 2021-02-15
Epoch.start_time is inconsistent with the spec


grahamgower:
The spec and reference implementation have done away with the `start_time` property for an epoch. This certainly makes sense for the YAML input, and the data model itself. However, once a YAML file is parsed into a Graph, it's rather useful to be able to inspect the start time of an epoch, without having to look outside the epoch object. Just like the `Deme` class has useful `start_time` and `end_time` properties (as opposed to attributes), so too would it be great to make `Epoch.start_time` a property, rather than an attribute. But if we remove the attribute from the epoch, then it no longer has this information. Should we include a back reference from the Epoch to the Deme it applies to, and use that? Or should we really just remove `Epoch.start_time` and look at the end_time of the previous epoch when we need this info?

apragsdale:
I agree with you Graham that it would be good to be able to grab `Epoch.start_time` directly, even if it isn't explicitly allowed in the specification in the YAML. But if internally, if we ask for `Epoch.start_time` it can just grab the previous epoch's end time or the deme start time if it's the first epoch, maybe that's good enough? I could see that being a bit of a logic pain though. I don't know what the cleanest or most principled way of doing that would be here.

jeromekelleher:
Yes, this is a tricky one - I can see arguments in both directions. Seems to me it's a natural thing to fill out during a ``parse`` operation, where we validate the graph and fill in the implicitly defined information according to the spec. It's not in the spec because it's redundant, but there's nothing stopping us from adding useful additional properties if we like.

However, it should be an error if someone tries to fill it in themselves. Which makes me think we should have a separate graph-building API, rather than composing the actual objects we define...

grahamgower:
> Which makes me think we should have a separate graph-building API, rather than composing the actual objects we define...

I've thought about doing this a couple of times, but the simplcity and elegance of using the API for building has remained very attractive. This works well, excepting some very minor inconsistencies (such as the present issue).

In #188, I've kept `start_time` it in the python `Epoch` class, and it is filled during parsing as you say @jeromekelleher (checking for inconsistencies as usual). It will be an error to specify in the input YAML file once #184 is adressed.

grahamgower:
Closed by #209.

PR #209, opened 2021-02-18
Add Builder API.

grahamgower:
Initial sketch of a possible Builder/Factory-style API, as suggested in #208. The code here is just for discussion and has not been tested. But I think the `Builder` class is fairly complete, and we can offload most of the error checking to the `resolve()` stage. I'm luke-warm on the DemeProxy helper class. Thoughts?

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1,5 +1,4 @@
-import typing
-from typing import List, Union, Optional, Dict
+from typing import List, Union, Optional, Dict, Mapping, MutableMapping, Any
 import itertools
 import math
 import numbers
@@ -67,27 +66,6 @@ def valid_deme_id(self, attribute, value):
         )
 
 
-def isclose(
-    a: Optional[Number],
-    b: Optional[Number],
-    *,
-    rel_tol=_ISCLOSE_REL_TOL,
-    abs_tol=_ISCLOSE_ABS_TOL,
-) -> bool:
-    """
-    Wrapper around math.isclose() that handles None.
-    """
-    if None in (a, b):
-        return (a,) == (b,)
-    else:
-        return math.isclose(
-            typing.cast(float, a),
-            typing.cast(float, b),
-            rel_tol=rel_tol,
-            abs_tol=abs_tol,
-        )
-
-
 def isclose_deme_proportions(
     a_ids: List[ID],
     a_proportions: List[Proportion],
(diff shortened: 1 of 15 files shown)


jeromekelleher:
LGTM. Some high-level thoughts:

1. We should provide an API for setting defaults, mirroring the spec
2. It might be worth looking at refactoring the resolution logic out of the Graph class, and into here (or some intermediate layer). Not necessary at all for a first pass, though.
3. The DemeProxy seems fine to me; we can also add a ``Builder.add_epoch(deme_id, ...)``. Both will be useful in different contexts, I think.

grahamgower:
There's still more test coverage needed, and few cleanups are warranted after all the churn, but this should give a pretty good indication of where things are headed.

Actually, it seems that separating the Builder from the Graph isn't wholly necessary. The nicest restructuring here is enabled by switching from `add_deme(..., epochs: List[demes.Epoch] = ...)` to `add_deme(..., epochs: List[dict] = ...)`. This change alone fixes most the awkwardness pointed out in #201.

grahamgower:
Hey @jeromekelleher, I think this is more or less done, so it would be great if you could take another look. There are a bunch of changes that went in here because it didn't make much sense to do a refactor, only to refactor again to conform with the spec. So after this PR, we should have parity with the reference implementation.

demes/demes.py
```diff
@@ -991,50 +1030,108 @@ class Deme:
...
         for proportion in self.proportions:
             unit_interval(self, attribute, proportion)
             positive(self, attribute, proportion)
 
     @epochs.validator
     def _check_epochs(self, _attribute, _value):
         # every deme must have at least one epoch
-        if len(self.epochs) == 0:
-            raise ValueError("Demes must be defined with at least one epoch")
+        # if len(self.epochs) == 0:
```
jeromekelleher, line 1053: Delete?

demes/demes.py
```diff
@@ -1937,3 +2036,196 @@ def collapse_demes(pairs):
...
+        :param float generation_time: The generation time of demes, in units given
+            by the ``time_units`` parameter. Concretely, dividing all times
+            by ``generation_time`` will convert the graph to have time
+            units in generations.  If ``generation_time`` is ``None``, the units
+            are assumed to be in generations already.
+        :param doi: If the graph describes a published demography, the DOI(s)
+            should be be given here as a list.
+        :vartype doi: list of str
+        """
+        self.data: MutableMapping[str, Any] = dict(time_units=time_units)
```
jeromekelleher, line 2082: Maybe add the lists here, to simplify things a bit later? E.g, ``dict(time_units=time_units, demes=[], migrations=[]...)``

jeromekelleher (approved):
I like it. Making the Builder responsible for working with the data model exclusively and defining its output in terms of the dict representation is a great idea. This makes the role of the class very well defined. We can make a long-term stable API on top of this.

Happy to merge whenever you are.

demes/demes.py
grahamgower: I had that at some point, but figured that having the explicit check and list construction inside `add_deme()` provides slightly greater robustness/flexibility in the case when someone uses `Builder.fromdict()`.

Issue #170, opened 2020-12-04
rename list_demographic_events()


grahamgower:
The method is classifying the types of ancestor/descendant relationships in the graph (pulse, split, branch, merger, or admix), so it would be nice if it had a name in line with that. Suggestions?

apragsdale:
They're all ancestor/descendant relationships, except for `pulse` events... though it is also an instantaneous/discrete event. Some ideas:

 - `discrete_demographic_events()`, since it lists all discrete mass migration type events but not continuous migration
 - or even just `discrete_events()` for conciseness?
 - `mass_migrations()` - kind of in line with how those event types are specified in msprime or fwdpy11

grahamgower:
Closed by #213.

Issue #99, opened 2020-11-17
test coverage should be independent of converter tests


grahamgower:
The `demes.convert.*` functions for msprime/stdpopsim/moments will likely be removed. So we want to ensure our test coverage is not dependent on the tests for these conversion functions. Related to #68.

apragsdale:
I've started working on a larger update release of `moments` that should be ready to go in the next few weeks, and including `demes` support natively within `moments` is on the list of tasks there. So I'm almost ready to remove the moments functions from `demes.convert`, and in any case it will need some updating because the `moments` API is getting streamlined quite a bit.

For `fwdpy11` as well, those conversion functions have been set up and tested elsewhere.

What do you think needs to take place within `stdpopsim` and `msprime` to get the `demes` support within those libraries instead of kept in here?

grahamgower:
Oh, nice! I think we want to have at least a `0.1` release before including `demes` as a dependency anywhere. I'd imagine it won't be an enourmous amount of work to get it into msprime (but we coule easily wait for after msprime 1.0?). It might be more effort in stdpopsim, because there's not much point in doing that without deprecating the current way of defining demographic models (possibly a large task).

jeromekelleher:
Once there's a release of demes, we can easily add the code we have here into the msprime test suite. Once we're ready to make demes a dependency (either in 1.0 or 1.1) we could promote the code to the main library.

grahamgower:
Test coverage is now in pretty good shape. The only thing we're relying on the converter tests for is `Graph.successors`, `Graph.predecessors`, and `Graph.list_demographic_events()`, which are only used in the moments conversion code.

apragsdale:
> The only thing we're relying on the converter tests for is `Graph.successors`, `Graph.predecessors`, and `Graph.list_demographic_events()`, which are only used in the moments conversion code.

We can discuss today in our call, but I think we're ready to remove the `moments` conversion code in `demes`. I've already migrated a lot of it over, getting ready for a `moments` release once `demes` is out and `pip`-installable. I think the only use to keep it for now is as an example for someone else to get some inspiration, like maybe Ryan with `dadi` (but we can always point developers to other repos where its supported).

We still want to keep these `successors`, `predecessors`, etc functions, so we'll want to explicitly test them elsewhere.

grahamgower:
Closed by #213.

PR #213, opened 2021-02-22
successors, predecessors, and discrete demographic events

### LINKED ISSUE #173: make successors and predecessors functions instead of properties?
I've been thinking about graph operations today, and it seems like we can convert to `networkx` via:
```python
import networkx as nx
import demes
g = demes.load("some-graph.yml")
gs = nx.from_dict_of_lists(g.successors, create_using=nx.DiGraph)
gp = nx.from_dict_of_lists(g.predecessors, create_using=nx.DiGraph)
```
This is very nice, and gets users immediate access to DAG traversals, transitive closure, etc. from `networkx`. There are two things worth noting here though.
 1. The networkx graphs give a time-agnostic view of the deme relationships.
 2. Pulses and migrations are not edges in these graphs.

If we made successors/predecessors functions, then we could add argument(s) to include pulses/migrations. I'm not sure about the forwards-in-time view (successors) with pulses/migrations added, but the backwards-in-time view (predecessors) of the ancestry is probably very useful. Combining this with a `prune()` operation (which uses/modifies time fields), should be very powerful.

grahamgower:
* Successors and predecessors are now functions, rather than properties.
* list_demographic_events() has been renamed to discrete_demographic_events().
* Tests added for each of the above.
* Split, Branch, Merge, Admix added to docs/api.rst.

Closes #173 #170 #99.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1,4 +1,4 @@
-from typing import List, Union, Optional, Dict, Mapping, MutableMapping, Any
+from typing import List, Union, Optional, Dict, Mapping, MutableMapping, Any, Set
 import itertools
 import math
 import numbers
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):
LGTW(FWIW)

grahamgower:
Thanks @jeromekelleher. I think I'll merge this while we've got some momentum, and if @apragsdale has any issues we can deal with that in a follow up issue.

PR #218, opened 2021-02-23
Fix Graph.assert_close() for pathological migration lists.

### LINKED ISSUE #212: assert_close() is sensitive to ordering of nested lists
The following test case fails, illustrating the problem for the migrations lists.
```python
import demes

b1 = demes.Builder(defaults=dict(epoch=dict(start_size=1)))
b1.add_deme("a")
b1.add_deme("b")
b1.add_deme("aa")
b1.add_migration(demes=["a", "b"], rate=0.1)
b1.add_migration(demes=["b", "aa"], rate=0.1)
g1 = b1.resolve()

b2 = demes.Builder(defaults=dict(epoch=dict(start_size=1)))
b2.add_deme("a")
b2.add_deme("b")
b2.add_deme("aa")
b2.add_migration(demes=["aa", "b"], rate=0.1)
b2.add_migration(demes=["b", "a"], rate=0.1)
g2 = b2.resolve()

g1.assert_close(g2)
```

This happens because `[sort_attributes(a) for a in sorted(x)] != sorted(sort_attributes(a) for a in x)`. We're doing the former, but need to instead do the latter. ~Note that the same problem will exist for ancestors/proportions lists within the list of demes, which is annoying to solve (because the ancestors and proportions lists must be sorted synchronously).~ EDIT: there's no problem with ancestors/proportions in the demes list, because they are first sorted by ID and IDs are unique, giving a well-defined ordering that ignores ancestor ordering. So it's just the `Graph.migrations` that have this problem.

grahamgower:
Closes #212.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1309,10 +1309,10 @@ class Graph:
             comparisons. See documentation for :func:`math.isclose`.
         """
 
-        def assert_sorted_eq(aa, bb, *, rel_tol, abs_tol, name, key=None):
+        def assert_sorted_eq(aa, bb, *, rel_tol, abs_tol, name):
             # Order-agnostic equality check.
             assert len(aa) == len(bb)
-            for (a, b) in zip(sorted(aa, key=key), sorted(bb, key=key)):
+            for (a, b) in zip(sorted(aa), sorted(bb)):
                 try:
                     a.assert_close(b, rel_tol=rel_tol, abs_tol=abs_tol)
                 except AssertionError as e:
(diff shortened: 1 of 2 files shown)


tests/__init__.py
```diff
@@ -200,7 +201,7 @@ def test_something(self, g):
             time_hi = min(
                 b.data["demes"][j]["start_time"], b.data["demes"][k]["start_time"]
             )
-            if time_hi - time_lo < 1e308:
```
jeromekelleher, line 203: Haha, quite a big difference here! Perhaps a sign should be using a constant here, e.g. from  numpy.finfo

jeromekelleher: Comparing to a constant is pretty weak, we should be using math.isclose or something anyway.

jeromekelleher (approved):
Looks good, but doing approx floating point comparisons by hand is error prone.

tests/__init__.py
grahamgower: The problem here is that if we generate values for `time_hi` and `time_lo`, then later draw more numbers in the open interval `(time_hi, time_lo)`. So if `time_hi` and `time_lo` are too close, hypothesis gives an error that there aren't any floating point numbers within that range. Any idea which particular floating point constant is appropriate here? I *think* the spacing between floating point numbers differs across the range, so I don't know if there is one number that applies. Some small delta should be ok though (and maybe that's bigger than `1e-308`)?

jeromekelleher: Hmm, what about  if ``time_hi == np.nextafter(time_lo)``? That tells you there's no gap between them. https://numpy.org/doc/stable/reference/generated/numpy.nextafter.html#numpy.nextafter

grahamgower: Ok, I'll rip that commit out and do it properly as a separate PR. It fails on my workstation, but I haven't seen it fail in CI yet.

PR #222, opened 2021-02-24
Fix defaults when using Builder, and add more tests.

### LINKED ISSUE #214: all args to Builder.add_migration() and Builder.add_pulse() should be optional
All args to `Builder.add_migration()` and `Builder.add_pulse()` should be optional. This is needed because it's possible to have these specified in the `defaults` section. But I also don't like it, because it's then easier to make mistakes (and probably most people won't be using migration or pulse defaults).

grahamgower:
The tests here also uncovered a small issue with migration defaults.
If the migration default in question is "demes", then the user must
override to obtain an asymmetric migration. Similarly, if the default is
the "source" and/or "dest", these must be overriden to obtain a
symmetric migration. With the changes here, overrides of 'None' can now
be used with the Builder API.

Closes #214.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1941,36 +1941,31 @@ class Graph:
             insert_defaults(migration_data, migration_defaults)
             if "rate" not in migration_data:
                 raise KeyError(f"migration[{i}]: required field 'rate' not found")
+            demes = migration_data.pop("demes", None)
+            source = migration_data.pop("source", None)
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):

grahamgower:
Hmm.. I was expecting to also see a codecov report. Not sure what's going on there.

PR #225, opened 2021-02-25
Fix Graph.in_generations() to convert the deme start_time.

### LINKED ISSUE #224: `in_generations` does not convert `deme[id].start_time`
If we have a deme graph in years, for example, and call `g_gen = g.in_generations()`, the resulting deme graph `g_gen` does not properly convert deme start times.

For example:

```Python
import demes
file_in = "examples/gutenkunst_ooa.yml"

g = demes.load(file_in)

print(g["AMH"].start_time)
print(g["AMH"].epochs[0].start_time)

g_gen = g.in_generations()

print(g_gen["AMH"].start_time)
print(g_gen["AMH"].epochs[0].start_time)

# note that there is no issue with end time, because
# of the end_time function grabbing the last epoch's end time
print(g_gen["AMH"].end_time)
print(g_gen["AMH"].epochs[0].end_time)
```

I _think_ this is just a quick fix of updating the `in_generations` function to also set `deme.start_time /= generation_time`, but @grahamgower maybe there is something more subtle going on?

grahamgower:
Also fixes ``Deme.assert_close()`` to check the ``start_time``.

Closes #224.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,17 @@
 ********************
+0.1.0a3 - 2021-02-25
+********************
+
+**Bug fixes**:
+
+- Fix ``Graph.in_generations()`` to also convert the ``Deme.start_time`` field.
+  Thanks to :user:`apragsdale` for reporting the problem.
+  (:user:`grahamgower`, :issue:`224`, :pr:`225`).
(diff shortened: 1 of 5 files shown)


jeromekelleher (approved):
LGTM

PR #230, opened 2021-02-26
Remove the tutorial; replace with a quickstart section.

grahamgower:
The tutorial content is out of date with respect to this year's
spec changes and the YAML files there were unfortunately missed
during updates. Also, the tutorial now belongs in the spec repository.


demes/demes.py
```diff
@@ -2103,7 +2103,8 @@ class Builder:
 
     :ivar dict data: The data dictionary of the graph's current state.
         The objects nested within this dictionary follow Demes' data model,
-        as described in the :ref:`schema <sec_schema>`.
+        as described in the
+        `schema <https://popsim-consortium.github.io/demes-spec-docs/main/spec.html>`_.
```
jeromekelleher, line 2107: Better to use intersphinx for this - it's a sphinx site, right, so it has an objects.inv etc.

jeromekelleher (commented):
LGTM, but we should use intersphinx.

demes/demes.py
grahamgower: Oh cool, thanks. I've made about 100 tiny mistakes here though... maybe I should just quit and go have a beer instead.

jeromekelleher: If you have to ask the question, it's definitely time for beer. Enjoy!

PR #238, opened 2021-03-05
Include py.typed file in distribution files.

### LINKED ISSUE #237: mypy doesn't find demes' type hints
```python
# a.py
import sys
import demes

def foo(graph: demes.Graph):
    pass

if len(sys.argv) != 2:
    print("usage: {sys.argv[0]} graph.yaml")
    exit(1)

graph = demes.load(sys.argv[1])
foo(graph)
```

```
$ mypy a.py
a.py:2: error: Skipping analyzing 'demes': found module but no type hints or library stubs
a.py:2: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

Mypy docs suggest we need to add a `py.typed` file.
https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages

grahamgower:
Closes #237.


grahamgower:
It looks like we just need to add this empty file. It seems `setuptools_scm` takes care of getting it in to the distribution files.

PR #241, opened 2021-03-18
Fix migration.end_time in convert.from_msprime.

grahamgower:
I think this is right now. This stuff all needs updating for msprime 1.0 anyway...


grahamgower:
Yep, but I just noticed this (and needed it to fix a model I'm working with). Figured it was better uploaded here instead of me forgetting about it in 10 minutes.

Issue #216, opened 2021-02-23
include SymmetricMigration in the hypothesis testing strategy


grahamgower:
The random Graph generator code in `tests/__init__.py:graphs()` currently only generates graphs with asymmetric migrations.

grahamgower:
The generated migrations must be careful not to produce migration matrix rows that sum>1. See #244 and #249.

grahamgower:
The relevant bits here were taken care of in #268.

PR #249, opened 2021-03-19
Check sum of migration rates entering a deme are <= 1.

### LINKED ISSUE #244: enforce an upper bound for the migration rate
Once we know what units we're using, we should be able to figure out the upper bound. https://github.com/popsim-consortium/demes-spec/issues/21

grahamgower:
Closes #244.

Probably needs more tests. I'm also not sure of the best way to avoid hitting this in the hypothesis strategy for generating graphs (but I didn't get any failures yet).

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -498,7 +498,7 @@ class Pulse:
         validator=[attr.validators.instance_of(str), valid_deme_name]
     )
     dest: Name = attr.ib(validator=[attr.validators.instance_of(str), valid_deme_name])
-    time: Time = attr.ib(validator=[int_or_float, non_negative, finite])
+    time: Time = attr.ib(validator=[int_or_float, positive, finite])
     proportion: Proportion = attr.ib(validator=[int_or_float, unit_interval])
 
     def __attrs_post_init__(self):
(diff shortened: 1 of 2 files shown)


demes/demes.py
```diff
@@ -1688,6 +1696,51 @@ def _add_pulse(self, *, source, dest, proportion, time) -> Pulse:
...
+    def _check_migration_rates(self):
+        """
+        Check that the sum of migration ingress rates doesn't exceed 1 for any
+        deme in any interval of time.
+        """
+        start_time = math.inf
+        mm_list, end_times = self._migration_matrices()
+        for migration_matrix, end_time in zip(mm_list, end_times):
+            for j, row in enumerate(migration_matrix):
+                if sum(row) > 1:
```
jeromekelleher, line 1736: Maybe add a ``and not math.isclose(row_sum, 1)`` to allow for the case when the sum is 1.0000000001 (which happens).

jeromekelleher (approved):
LGTM

grahamgower:
I'm going to merge this without fixing the hypothesis strategy. The migration-generating part of that code needs to be updated for symmetric migrations anyway (#216), so I'll add a note about the sum<=1 restriction over there.

PR #256, opened 2021-03-24
Convert docs to jupyter-book.

grahamgower:
I probably missed something, but this should take care of most things.


demes/demes.py
```diff
@@ -1877,7 +1870,7 @@ def discrete_demographic_events(self) -> Dict[str, List[Any]]:
             )
         return demo_events
 
-    def in_generations(self):
+    def in_generations(self) -> "Graph":
```
jeromekelleher, line 1873: Maybe use the ``from __future__`` thing so you can write it without the quotes? Or is that only available in 3.7+?

docs/_config.yml
```diff
@@ -0,0 +1,48 @@
...
+    - sphinx.ext.intersphinx
+    - sphinx_issues
+
+    config:
+      issues_github_path: popsim-consortium/demes-python
+      todo_include_todos: true
+      intersphinx_mapping:
+        python: ["https://docs.python.org/3/", null]
+        spec: ["https://popsim-consortium.github.io/demes-spec-docs/main/", null]
+        stdpopsim: ["https://stdpopsim.readthedocs.io/en/stable", null]
```
jeromekelleher, line 41: Do we need stdpopsim and msprime intersphinx mappings? We shouldn't really be depending on their APIs, right?

docs/changelog.md
```diff
@@ -1,5 +1,5 @@
-=========
-Changelog
-=========
+# Changelog
 
+```{eval-rst}
 .. include:: ../CHANGELOG.rst
```
jeromekelleher, line 4: Might be as easy update the CHANGELOG to markdown now. It'll make it easier to work with in GitHub.

What I did for msprime is put in a soft link to the CHANGELOG.md file in the project root here. Works fine.

jeromekelleher (approved):
Fantastic! Minor comments.

demes/demes.py
grahamgower: Unfortunately that was introduced in 3.7. BTW, that future becomes the default behaviour in 3.10.

docs/_config.yml
grahamgower: Just for the `demes.convert` functions. How about I just remove that docs page now though?

jeromekelleher: Yes, I think so. There's no point in documenting something we expect to remove immediately and don't want people to use.

PR #268, opened 2021-04-05
Resolve symmetric migrations into pairs of asymmetric migrations.

### LINKED ISSUE #263: nail down semantics for many-deme symmetric migrations
The semantics for many-deme symmetric migrations are not clear. https://github.com/popsim-consortium/demes-spec/issues/55. I'm leaning towards making this an error as Jerome suggests in the spec issue. There are probably some tricky details to be worked out here, with regard to the `Graph.asdict_simplified()` code --- I yanked out the migration simplification code again a few weeks back (sorry Aaron!), in part because it didn't fit with the splitting into two migration classes, but also because of the lack of clarity here. It would be really great to have the migration simplification code back.

grahamgower:
Also restores the migration simplification code that was removed
in commit e2a7e05 (#209).

Closes #263.

### DIFF
--- a/tests/test_demes.py
+++ b/tests/test_demes.py
@@ -3,6 +3,7 @@ import copy
 import pathlib
 import math
 import typing
+import random
 
 import pytest
 import hypothesis as hyp
@@ -11,7 +12,6 @@ from demes import (
     Builder,
     Epoch,
     AsymmetricMigration,
-    SymmetricMigration,
     Pulse,
     Deme,
     Graph,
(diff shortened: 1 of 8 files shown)


tests/test_demes.py
```diff
@@ -11,7 +12,6 @@
     Builder,
     Epoch,
     AsymmetricMigration,
```
apragsdale, line 14: If we only have AsymmetricMigration as the internal continuous migration class, should we call it simply Migration? Or leave it as AsymmetricMigration so that we leave open the option of reintroducing SymmetricMigration?

apragsdale (commented):
This looks great to me! Lots of fiddling, but I think it will end up simplifying things again quite a bit. I also like that it gets the migration simplification back in there when using `asdict_simplified` :)

tests/test_demes.py
grahamgower: Yeah, good question. I left it as `AsymmetricMigration` initially because it was easiest at the first stage of refactoring. But having this be explicit might help reduce confusion (at the expense of verbosity). And as you say, its not implausible that we'd introduce something called SymmetricMigration again in the future. But I don't have a strong opinion on what this should be called.

apragsdale: More of a question than suggestion that we should change it. I'm happy to keep it as AsymmetricMigration, as it does reduce any possible confusion (and also allows us to introduce a SymmetricMigration if we are compelled to in the future again). Most users will still just see the `add_migration` function in the Builder, and won't even know how it is being represented behind the scene, I imagine.

apragsdale (approved):

PR #274, opened 2021-04-15
Fix in_generations() with time_units=generations and generation_time.

### LINKED ISSUE #273: `Graph.in_generations()` assumes `time_units` are not generations if `generation_time` is given.
This is wrong, because it means the times get converted to nonsense values for, e.g. `examples/zigzag.yml` (which has `time_units=generations; generation_time=30`).

grahamgower:
Closes #273


jeromekelleher (approved):
LGTM

PR #278, opened 2021-04-16
Clean up size_function semantics.

### LINKED ISSUE #262: nail down semantics for `size_function`
The semantics for `size_function` are not nailed down. See https://github.com/popsim-consortium/demes-spec/issues/34. We're currently allowing any value, but treating "constant" and "exponential" specially (during consistency checks, and during `Graph.asdict_simplified()`). When not specified the defaults are currently:
```python
        if size_function is None:
            if start_size == end_size:
                size_function = "constant"
            else:
                size_function = "exponential"
```
As far as I'm concerned, it would be easiest if we only supported these two options, as then basically all downstream software can do something sensible. But easiest is not necessarily best. Do we want to allow any string here? Do we want to allow other values, such as "linear"? Do we permit downstream software to ignore values it doesn't understand? We currently have a warning in the docs `Warning: Downstream applications might not understand the size_function provided.`, but I find this somewhat unsatisfying --- clearly it would be best if they did.

grahamgower:
Closes #262.


demes/demes.py
```diff
@@ -1012,12 +1029,8 @@ def _add_epoch(
             if end_size is None:
                 end_size = start_size
 
-        # XXX: use math.isclose()?
         if size_function is None:
-            if start_size == end_size:
```
apragsdale, line 1017: We still want it to default to "constant" if the start and end sizes are the same, right?

apragsdale (commented):
This is looking good - I like the documentation you've added. Just one question about always defaulting to "exponential" here, but maybe I'm missing something..

demes/demes.py
grahamgower: That's definitely an option. I just changed the behaviour to match what Jerome suggested in #262. Defaulting to `exponential` is *slightly* simpler.

apragsdale: I think I had misunderstood Jerome's suggestion before. I'm imagining software that only handle piecewise constant, so they check the size function of each epoch in a loaded graph, but if the default is to exponential, they'd fail even when start size == end size.

grahamgower: I suppose. Did you have any specific software in mind here? Surely software which can only handle piecewise-constant models are able to partition each exponential epoch into `m` piecewise-constant subepochs?

apragsdale: I'm thinking of PSMC/stairway plot type methods. I don't think we'd want to ask them to partition an epoch that is actually already constant but having a size function of "exponential". There are other methods that are more expensive to compute transitions for an exponential epoch as opposed to a constant size epoch, so they'd be interested in seeing the "constant" value if it actually is a constant size.

grahamgower: The change here only affects the default value when reading a YAML file that has no `size_function` explicitly given. Any inference method that produces piecewise-constant epochs can specify `defaults: {epoch: {size_function: constant}}` in the YAML they output.

apragsdale: Right - but in `moments` for example, if I read in a YAML file and want to compute the SFS, I create transition matrices for given sizes in each epoch. For constant sizes, these are computed once. But they need to be recomputed every time step of integration if the sizes are non-constant functions, and `moments` reads the `size_function` field to determine this.

Pretty sure users and other developers would also be confused to see epochs where the start and end times are equal, and then see that the size function defaulted to exponential when constant is an option.

grahamgower: Ok, I'm happy to change this. Just wanted to talk it through in case I was missing something obvious. :)

Would you be satisfied with defaulting to "constant" `if not math.isclose(start_size, end_size)`, rather than `if start_size != end_size`? The spec wording could be something like: if size_function is not specified, implementations are expected to set the size_function to "constant" when start_size and end_size are the same (to within an implementation-defined numerical tolerance), and "exponential" otherwise.

apragsdale: Ok, no problem! Yeah, I think that's reasonable. Do you mean something like `if math.isclose(start_size, end_size)` -> "constant", else "exponential"?

grahamgower: > Ok, no problem! Yeah, I think that's reasonable. Do you mean something like `if math.isclose(start_size, end_size)` -> "constant", else "exponential"?

Ughh.. yes. Sorry!

molpopgen: > I think I had misunderstood Jerome's suggestion before. I'm imagining software that only handle piecewise constant, so they check the size function of each epoch in a loaded graph, but if the default is to exponential, they'd fail even when start size == end size.

I understood that comment to mean "exponential if start != end".

molpopgen: I think this is looking good!

demes/demes.py
```diff
@@ -191,7 +206,9 @@ def __attrs_post_init__(self):
         # XXX: these tests should use math.isclose()
         if math.isinf(self.start_time) and self.start_size != self.end_size:
             raise ValueError("if start time is inf, must be a constant size epoch")
-        if self.size_function == "constant" and self.start_size != self.end_size:
+        if self.size_function == "constant" and not math.isclose(
```
jeromekelleher, line 209: I think we should require them to be *exactly* equal if "constant" is the size function - it's semantically a bit weird otherwise

jeromekelleher (commented):
This looks good. Reading through the thread, here's my suggestion:

1. If start_size == end_size and no size function is specified, default to "constant"
2. If start_size != end_size and size_function == "constant" raise an error

That is, hard equality rather than math.isclose. It's hard to imagine cases where the values would differ if the size function really was constant and it's not an error.

demes/demes.py
grahamgower: Ok, I've updated to do exact equality checking (in both locations).

PR #281, opened 2021-04-21
Remove the console script entry point.

grahamgower:
I think this ought to do the trick @jeromekelleher.


jeromekelleher (approved):
LGTM

jeromekelleher:
We should probably remove the ``__main__`` entry point as well though? No point in having it if it's not documented/used.

grahamgower:
Ok, I removed the `__main__.py` too. But I've personally found this quite useful to validate yaml files that I write for examples. I guess `python -c "import demes; demes.load('foo.yaml')"` isn't much for typing though. I expect we will want some kind of CLI in the future. I imagine printing the fully resolved model might be useful for users if they're uncertain how the defaults will be applied to their model. It might also be useful to have CLI converters?

jeromekelleher:
Yeah, definitely agree it'd be nice to have a CLI. Lets open an issue to brainstorm

PR #294, opened 2021-05-11
Hypothesis strategy cleanups.

### LINKED ISSUE #217: export the hypothesis strategy for generating random Graph objects
This is likely very useful for downstream packages. But there are some quirks that need to be dealt with:
https://hypothesis.readthedocs.io/en/latest/strategies.html#registering-strategies-via-setuptools-entry-points

### LINKED ISSUE #292: hypothesis strategy generates corner cases we don't care about
Brought up when testing the msprime converter: https://github.com/tskit-dev/msprime/pull/1678.

Probably we should limit the strategy to generating sane deme sizes (i.e. `1 >= size > 1e80`), and sane event times (i.e. `1>= times_in_generations > 1e80`).

grahamgower:
It turns out that using the setuptools entry_points feature buys us basically nothing. It's far simpler to just put the hypothesis strategy into a submodule, and the tests in downstream apps can just import that submodule. Regular users that do `import demes` won't be importing this.

I've also restricted the range of the floating point numbers that are generated for deme sizes and event times. I may have overlooked something, but this works fine when used with the recently added msprime/demes tests. Furthermore, there's a new `min_deme_size` parameter to the graphs() strategy which ought to be useful to restrict sizes to be >=1 in cases where that is appropriate.

Closes #217.
Closes #292.

### DIFF
--- a/demes/hypothesis_strategies.py
+++ b/demes/hypothesis_strategies.py
@@ -1,12 +1,35 @@
 import math
 import itertools
 import collections
+import struct
 
-import numpy as np
 import hypothesis as hyp
 import hypothesis.strategies as st
 
 import demes
+
+
+def prec32(x):
+    """truncate x to the nearest single-precision floating point number"""
+    return struct.unpack("f", struct.pack("f", x))[0]
+
+
+# Limits for the floating point numbers we'll draw.
+#
+# We wish to be more restrictive with the allowable range than the limits
+# provided by floating-point types, to avoid doing arithmetic on numbers at
+# those floating point limits. Values near the limits are not useful for
+# demographic models in practice, so we don't want to generate models that
+# require applications to deal with floating point underflow and overflow.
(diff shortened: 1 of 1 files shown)


demes/hypothesis_strategies.py
```diff
@@ -1,14 +1,37 @@
...
+#
+# We wish to be more restrictive with the allowable range than the limits
+# provided by floating-point types, to avoid doing arithmetic on numbers at
+# those floating point limits. Values near the limits are not useful for
+# demographic models in practice, so we don't want to generate models that
+# require applications to deal with floating point underflow and overflow.
+# On the other hand, we also don't want to enforce artificial limits in the
+# Demes spec for things like time values or deme sizes.
+#
+# The numbers below are sufficiently convservative so as to avoid underflow
```
jeromekelleher, line 27: typo convservative

jeromekelleher (approved):
LGTM

Issue #296, opened 2021-05-31
Linear size function


noscode:
I have [added demes to GADMA](https://github.com/ctlab/GADMA/commit/911e8e672414038138525f0ccc4e309a8ed78e9a) to use it for the alternative representation of the demographic histories. I also used demesdraw for drawing. I have faced to one problem though. By default, GADMA uses three types of size functions: constant, exponential and linear. Unfortunately, demes does not support linear function right now and GADMA has to print warnings when it try to use demes for the demographic history with linear size change. It is not very big deal but it would be very cool to get support of linear functions.

I found some discussion here #262: additional size function support was planned in the future. What do you think about adding linear now? If you like I can implement it as then GADMA will be able to use demes for any history and I really like it to do so.

grahamgower:
> I have [added demes to GADMA](https://github.com/ctlab/GADMA/commit/911e8e672414038138525f0ccc4e309a8ed78e9a) to use it for the alternative representation of the demographic histories. I also used demesdraw for drawing.

Oh, wow! Cool!

> I have faced to one problem though. By default, GADMA uses three types of size functions: constant, exponential and linear. Unfortunately, demes does not support linear function right now and GADMA has to print warnings when it try to use demes for the demographic history with linear size change. It is not very big deal but it would be very cool to get support of linear functions.
> 
> I found some discussion here #262: additional size function support was planned in the future. What do you think about adding linear now? If you like I can implement it as then GADMA will be able to use demes for any history and I really like it to do so.

Yes, this sounds good. It would be great if you could add in the bits needed for `size_function == "linear"`. Probably this doesn't need much, just a few lines plus the tests.

I suppose we should keep the current behaviour where if size_function is not specified in a YAML, but start_size != end_size, then the size_function defaults to "exponential".

Any other thoughts @apragsdale?

noscode:
> I suppose we should keep the current behaviour where if size_function is not specified in a YAML, but start_size != end_size, then the size_function defaults to "exponential".

I agree, I was thinking about the same thing.

jeromekelleher:
This is great - we should update the spec also.

grahamgower:
Closed by #310.

PR #310, opened 2021-06-01
Add linear size function support

noscode:
Solves issue #296.
I have add `linear` as available size function in `demes`. I am not sure about the order of size functions in doc strings and in lists: I have put `linear` at the end as it is less popular than `constant` and `exponential`.
I also have created new example .yml file for demography with several linear size changes in order to check load and dump in tests.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -163,8 +163,8 @@ class Epoch:
         If ``start_size != end_size``, the population size changes
         monotonically between the start and end times.
     :ivar str size_function: The size change function. This is either
-        ``constant`` or ``exponential``, though it is possible that
-        additional values will be added in the future.
+        ``constant``, ``exponential`` or ``linear``, though it is possible
+        that additional values will be added in the future.
 
          * ``constant``: the deme's size does not change over the epoch.
          * ``exponential``: the deme's size changes exponentially from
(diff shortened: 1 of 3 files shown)


grahamgower (approved):
LGTM!

grahamgower:
Looks great, thanks @noscode. I'll give @apragsdale and @jeromekelleher a chance to take a look first, but I'm happy for this to be merged as it is.

noscode:
Cool!

> Looks great, thanks @noscode. I'll give @apragsdale and @jeromekelleher a chance to take a look first, but I'm happy for this to be merged as it is.

Yes, sure!

demes/demes.py
```diff
@@ -163,8 +163,8 @@ class Epoch:
         If ``start_size != end_size``, the population size changes
         monotonically between the start and end times.
     :ivar str size_function: The size change function. This is either
-        ``constant`` or ``exponential``, though it is possible that
-        additional values will be added in the future.
+        ``constant``, ``exponential`` or ``linear``, though it is possible
+        that additional values will be added in the future.
```
apragsdale, line 167: It would be good for the docs to state that if start_size != end_size then the default behavior is exponential, now that there are multiple options for changing population sizes.

apragsdale (commented):
Looks good to me as well. I'm happy to merge, just a short comment about docs

demes/demes.py
noscode: I can change an old warning that is below to the note that when  `start_size != end_size` then it is exponential size function.

demes/demes.py
```diff
@@ -177,12 +177,20 @@ class Epoch:
...
+               dt = (epoch.start_time - t) / epoch.time_span
+                N = epoch.start_size + (epoch.end_size - epoch.start_size) * dt
+
+        .. note::
 
-            Do not assume an exponentially changing size just because
-            ``start_size != end_size``. For forwards compatibility,
-            applications should always check the ``size_function``.
+            When ``start_size != end_size`` an exponentially changing size is
+            assumed.
```
grahamgower, line 193: I don't think this is the right place to mention this. The warning can probably just be removed, because now that we have "linear" as an option it should be self evident that when ``start_size != end_size`` the size function could be exponential or linear (and one must look at the `size_function` field to find out).

An exponentially changing size is only assumed when parsing a model (e.g. a from YAML, or from a demes.Builder()) and that model doesn't specify the `size_function`. But I think the docs here are mostly for people using an Epoch object, so this could potentially be confusing. A Epoch object *always* has the `size_function` set, and users should *always* check the `size_function` of the epoch to find out what to do when `start_size != end_size`.

I agree that we need to document the behaviour when reading in a model and the size_function isn't specified. But I think that should be in the spec instead.

jeromekelleher (approved):
LGTM, thanks @noscode!

grahamgower:
Thanks @noscode. Sorry about the back-and-forth.

noscode:
That is totally okay, happy to help!

PR #314, opened 2021-06-03
Add Deme.size_at() to compute the deme size at a given time.

### LINKED ISSUE #312: Compute size of deme at time t
Now that we have linear size functions too (#310), we should add a function that computes the size of a deme at time t (raising a ValueError if the deme doesn't exist at this time).

grahamgower:
Closes #312.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1143,6 +1143,36 @@ class Deme:
         """
         return self.start_time - self.end_time
 
+    def size_at(self, time: float) -> float:
+        """
+        Get the size of the deme at a given time.
+
+        :param float time: The time at which the size should be calculated.
+        :return: The deme size.
+        :rtype: float
+        """
+        for epoch in self.epochs:
+            if epoch.start_time >= time >= epoch.end_time:
+                break
+        else:
+            raise ValueError(
+                f"time {time} is outside deme {self.name}'s existence interval: "
+                f"({self.start_time}, {self.end_time}]"
+            )
+
+        if math.isclose(time, epoch.end_time) or epoch.size_function == "constant":
+            N = epoch.end_size
+        elif epoch.size_function == "exponential":
+            dt = (epoch.start_time - time) / epoch.time_span
+            r = math.log(epoch.end_size / epoch.start_size)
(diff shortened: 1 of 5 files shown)


demes/demes.py
```diff
@@ -1143,6 +1143,33 @@ def time_span(self):
...
+
+        :param float time: The time at which the size should be calculated.
+        :return: The deme size.
+        :rtype: float
+        """
+        for epoch in self.epochs:
+            if epoch.start_time >= time >= epoch.end_time:
+                break
+        else:
+            raise ValueError(f"deme {self.name} doesn't exist at time {time}")
```
apragsdale, line 1158: The error could give the deme's start and end time as a hint, perhaps? E.g. `ValueError(f"time {time} is outside of existence interval of deme {self.name}: ({deme.end_time}, {deme.start_time})")`

apragsdale (commented):
Looking good - just a small suggestion for the error message.

demes/demes.py
```diff
@@ -1143,6 +1143,36 @@ def time_span(self):
...
+    def size_at(self, time: float) -> float:
+        """
+        Get the size of the deme at a given time.
+
+        :param float time: The time at which the size should be calculated.
+        :return: The deme size.
+        :rtype: float
+        """
+        for epoch in self.epochs:
+            if epoch.start_time >= time >= epoch.end_time:
```
jeromekelleher, line 1155: Maybe it's the sleep deprivation, but I can't get my head around this condition - shouldn't it be ``epoch.start_time <= time < epoch.end_time``? As in, we break if the time is within this epoch's interval?

grahamgower: Two things: (1) `start_time > end_time`, so we must have `start_time > time >= end_time`. And (2) we also want to accept `time=math.inf`, so we check `start_time >= time >= end_time`. Given this is in a loop, and the epochs are time-sorted, the start_time = time condition is only hit for the first epoch, and only if the time is infinity.

jeromekelleher: Ack, I'm thinking backwards in time as usual, thanks, my bad!

PR #334, opened 2021-06-21
Fix typo

jeromekelleher:


grahamgower:
It seems the docs aren't building because references to the spec docs have disappeared.

PR #335, opened 2021-06-22
Add load_all()/dump_all() to support multidoc YAML.

### LINKED ISSUE #239: support multi-document YAML?
YAML permits multiple "documents" per file, using `---` dash separators. We should consider supporting this in the load/save API, to streamline how folks could use it for collections of Demes graphs (e.g. for inference).

grahamgower:
Closes #239.

### DIFF
--- a/tests/test_load_dump.py
+++ b/tests/test_load_dump.py
@@ -594,3 +594,78 @@ class TestLoadAndDump:
         assert data["migrations"][0]["start_time"] == "Infinity"
         g2 = demes.loads(json_str, format="json")
         g2.assert_close(g1)
+
+
+class TestMultiDocument:
+    @pytest.mark.parametrize("yaml_file", tests.example_files())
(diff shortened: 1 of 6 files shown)


tests/test_load_dump.py
```diff
@@ -594,3 +594,68 @@ def test_json_infinities_get_stringified(self):
         assert data["migrations"][0]["start_time"] == "Infinity"
         g2 = demes.loads(json_str, format="json")
         g2.assert_close(g1)
+
+
+class TestMultiDocument:
```
jeromekelleher, line 599: Can you check the empty file case, where we should return an empty genereator?

jeromekelleher (approved):
LGTM

tests/test_load_dump.py
grahamgower: Ah, good call! An empty genmerator sounds like the right thing to do.

PR #337, opened 2021-06-24
Use `__all__` to advertise public API.

### LINKED ISSUE #324: clean up the exported names
We should use `__all__` to only export names that are part of the public API. Also see mypy --strict output in #323, where the `Module has no attribute "foo"` suggests where we might consider explicit exports.

grahamgower:
Closes #324.

Note that non-public symbols are still accessible, and that some
applications, such as iPython, will ignore `__all__` when offering
tab-completion suggestions.

### DIFF
--- a/demes/__init__.py
+++ b/demes/__init__.py
@@ -1,5 +1,3 @@
-# flake8: noqa: F401
-
 __version__ = "undefined"
 try:
     from . import _version
@@ -31,3 +29,35 @@ from .load_dump import (
     dump_all,
 )
 from .ms import from_ms
+
+__all__ = [
+    "Builder",
+    "Epoch",
+    "AsymmetricMigration",
+    "Pulse",
+    "Deme",
+    "Graph",
+    "Split",
+    "Branch",
+    "Merge",
+    "Admix",
+    "load_asdict",
+    "loads_asdict",
+    "load",
+    "loads",
+    "load_all",
+    "dump",
+    "dumps",
+    "dump_all",
+    "from_ms",
+]
+
+
+# Override the symbols that are returned when calling dir(<module-name>).
+# https://www.python.org/dev/peps/pep-0562/
+# We do this because the Python REPL and IPython notebooks ignore __all__
+# when providing autocomplete suggestions. They instead rely on dir().
+# By not showing internal symbols in the dir() output, we reduce the chance
+# that users rely on non-public features.
+def __dir__():
+    return sorted(__all__)
--- a/demes/hypothesis_strategies.py
+++ b/demes/hypothesis_strategies.py
@@ -8,6 +8,12 @@ import hypothesis.strategies as st
 
 import demes
 
+__all__ = ["graphs"]
+
+
+def __dir__():
+    return sorted(__all__)
+
 
 def prec32(x):
     """truncate x to the nearest single-precision floating point number"""
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,6 +39,10 @@ extend-exclude = docs/_build
 # black-compatible settings
 max-line-length = 88
 extend-ignore = E203, W503
+# There's no way to ignore specific warnings in the files themselves.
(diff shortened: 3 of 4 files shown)


grahamgower:
Annoyingly, people who use python notebooks to explore libraries with tab completion (instead of reading documentation) will still see, and be able to access, non-public symbols. I believe the number of people who use notebooks like this is substantial.

```
$ ipython
Python 3.9.5 (default, May 24 2021, 12:50:35)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import demes

In [2]: demes.<TAB>
  Admix               Deme                dumps()             load()              loads()             Pulse
  AsymmetricMigration demes               Epoch               load_all()          loads_asdict()      Split
  Branch              dump()              from_ms()           load_asdict()       Merge
  Builder             dump_all()          Graph               load_dump           ms
In [2]: demes.ms.<TAB>
        Admixture                  build_graph()              demes_sorted_by_ancestry() from_ms()                  logging
        Any                        coerce_nargs()             Dict                       GrowthRateChange           Mapping
        argparse                   copy                       Event                      List                       math                       >
        attr                       demes                      finite()                   logger                     MigrationMatrixChange
In [2]: demes.ms.coerce_nargs
Out[2]: <function demes.ms.coerce_nargs(obj_creator, append=False)>
```

See https://github.com/ipython/ipykernel/issues/129

demes/load_dump.py
```diff
@@ -11,6 +11,17 @@
 
 import demes
```
jeromekelleher, line 12: Not clear to me why we'd have an ``__all__`` in implementation-private submodules. We woulnd't encourage users to use ``from demes.load_dump import *``, would we?

jeromekelleher (commented):
Looks good, but I don't know why we'd do this for anything other than the top-level __init__.py

Also, we need a new test module that does something like [this one](https://github.com/tskit-dev/msprime/blob/main/tests/test_imports_star.py)

(The strings in __all__ aren't resolved until run time - this is the only way to be sure they aremeaningul)

grahamgower:
IMO, this is more about what the public API is being advertised as, rather than caring about specific `import *` behaviour. I don't think anyone advocates `import *` anymore, except perhaps for legacy reasons (e.g. `from tkinter import *`). But `__all__` has become the de-facto standard list that editors use to provide tab-completion lists: `__all__` implicitly corresponds to the API that library authors are making public.

jeromekelleher:
Sure, I agree noone should actually be using this, but I still don't see why you'd want to add ``__all__`` lists to the internal modules. We actively want to hide the internal module structure from users (so that we can change it later, if we want to) not advertise it.

grahamgower:
In some sense you're right about not needing `__all__` in internal modules like `load_dump.py`. But unfortunately commonly used tools like the python REPL and ipython/notebooks will blindly tab-complete the names of internal modules. So I've updated the PR to go one step further and define `def __dir__(): sorted(__all__)`, because the python REPL and ipython notebooks don't tab-complete anything except what `dir(<module-name>)` shows.

A possible further step (which I don't want to take) is to delete the non-public symbols from the modules so that they can't be imported even if one knows the name (E.g. like this: https://github.com/tensorflow/tensorflow/blob/e9de087fa7f59c39bbe12ac2c83c5547c83f746c/tensorflow/python/util/all_util.py#L86).

demes/__init__.py
```diff
@@ -31,3 +31,30 @@
...
+    "load",
+    "loads",
+    "load_all",
+    "dump",
+    "dumps",
+    "dump_all",
+    "from_ms",
+]
+
+
```
jeromekelleher, line 58: Good to put in a comment to explain why we're doing this.

jeromekelleher (approved):
LGTM - just needs a test checking import * works, like I suggested, and something covering dir

PR #353, opened 2021-07-08
Simultaneous sources in pulse events

apragsdale:
WIP. First pass to update existing code and tests to handle sources and proportions as lists instead of str/value. See popsim-consortium/demes-spec#100 for discussion.

Need to expand testing of valid and invalid pulses with multiple sources.


grahamgower (commented):
LGTM @apragsdale. Just a couple of minor code nits.

jeromekelleher (approved):
Changes basically LGTM, but I haven't gone through the fine details.

apragsdale:
Thanks for the suggestions, @grahamgower. I think everything is looking reasonable here after rebasing from the merge of #362.

demes/ms.py
```diff
@@ -989,8 +989,8 @@ def get_growth_rate(epoch):
             pulse = deme_or_pulse
             num_demes += 1
             new_deme_id = num_demes
-            e1 = Split(pulse.time, deme_id[pulse.dest], 1 - pulse.proportion)
-            e2 = Join(pulse.time, new_deme_id, deme_id[pulse.source])
+            e1 = Split(pulse.time, deme_id[pulse.dest], 1 - pulse.proportions[0])
+            e2 = Join(pulse.time, new_deme_id, deme_id[pulse.sources[0]])
```
grahamgower, line 993: This looks like it will silently do the wrong thing when there are multiple pulse sources. I don't think it's too much work to support this, so feel free to do it if you're keen. Otherwise just raise a ValueError here if there are multiple sources, and we can open an issue to track it.

grahamgower:
Looks good @apragsdale. Sorry for giving you more work, with all the merge conflicts!

apragsdale:
No problem at all! Thanks @grahamgower. I'll squash and merge.

demes/ms.py
apragsdale: Yes good point. I'll place a value error and open an issue for this. Will be easier to address in a smaller fix than here.

PR #354, opened 2021-07-09
add to_ms()

### LINKED ISSUE #74: Conversion to/from ms format?
I'm guessing that a lot of people will want to be able to either convert their existing models to demes or to run demes models through simulators that support ms's demographic model description.

Should we provide ``from_ms()`` and ``to_ms()`` methods? (Or similarly named?)

The ``from_ms`` code wouldn't be too bad, as we could adapt the command line parsing code in [msprime](https://github.com/tskit-dev/msprime/blob/main/msprime/cli.py#L469). With a few small adjustments (i.e., raising ValueErrors rather than calling sys.exit) I think we could reuse the argparse code directly, which doesn't *have* to actually be used in the context of a CLI.

I've been thinking about doing this for the msprime 1.0 Demography API, but it seems to me it would be much more useful here.

Thoughts?

grahamgower:
Closes #74.

---
The tests here are somewhat circular, but I think this is as good as its going to get without actually running `ms`. And while I don't want to run `ms` in the CI tests, we almost certainly need to have some checks that do that.

Anyway, my brain hurts. So I'm stopping here.

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -1,0 +1,411 @@
+#!/usr/bin/env python3
+# Verifies the demes.to_ms() converter by comparing summary statistics against
+# msprime and moments. Tested with Python 3.9 on Linux.
+import abc
+import concurrent.futures
+import functools
+import subprocess
+import os
+
+import demes
+import demesdraw
+import msprime
+import moments
+import tsconvert
+import numpy as np
+import matplotlib
+
+# Use non-GUI backend, to avoid problems with multiprocessing.
+matplotlib.use("Agg")
+import matplotlib.pyplot as plt  # noqa: E402
+from matplotlib.backends.backend_pdf import PdfPages  # noqa: E402
+
+NUM_PROCS = os.cpu_count()
+NUM_REPLICATES = 100_000
+REPS_PER_BATCH = 5_000
+assert NUM_REPLICATES % REPS_PER_BATCH == 0
+# ms isn't so useful for testing, as the trees' branch lengths are printed
+# to only 3 decimal places, which is insufficient to build a tree sequence.
+# We just want to check demes' ms output is implemented correctly, so mspms
+# is a convenient alternative.
+MS_COMMAND = "mspms"
+
+
+class Simulator(abc.ABC):
+    """Abstract base class for simulators."""
+
+    def __init__(
+        self, *, graph: demes.Graph, samples: dict[str, int], num_replicates=None
+    ):
+        self.graph = graph
+        self.samples = samples
+        if num_replicates is None:
+            num_replicates = 1
+        self.num_replicates = num_replicates
+
+    def tmrca(self):
+        """Get vector of tmrcas, one for each replicate."""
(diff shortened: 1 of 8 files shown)


jeromekelleher:
Great to have this - haven't time to review properly right now though.

I guess we should create an issue to track creating a validation script, like msprime's verification.py? We could run a few representative models though ms and msprime, and see if we get the same distributions of coalescent times etc. Probably the only way to really be sure we're getting this stuff right. There's no point in running ms as part of CI.

grahamgower:
I'm having a bit of trouble with the verification scripts. The ms output looks fine, but the msprime output does not. I assume there's an issues scaling thngs so that the units all match, so I'd appreciate some assistance on this!
verification

verification.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+class SimMsprime(Simulator):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        demog = msprime.Demography.from_demes(self.graph)
+        ts = msprime.sim_ancestry(
+            demography=demog,
+            samples=self.samples,
+            ploidy=1,
```
jeromekelleher, line 119: This will make things off by a factor of 2, won't it? Probably simpler to run in diploid mode, but state your samples are haploid using the SampleSet notation: https://tskit.dev/msprime/docs/latest/ancestry.html#sec-ancestry-ploidy

jeromekelleher (commented):
spotted one probable source of error anyway

verification.py
grahamgower: Thanks, that seems to do the trick. But I'm very confused now. Shouldn't deme `a` and `b` both be affected? Why does matter which deme is being sampled from?

jeromekelleher: I've no idea, I didn't think about the details, I just scanned for obvious problems :smile:

grahamgower: I'm actually a little concerned about this, because it suggests that changing the ploidy between 1/2 can result in scaling of time that *isn't* different by a factor of 2. If you look at the plots above, the stats look fine when sampling from `a`, and the stats for `b` are off by some factor that is clearly not 2.

grahamgower: I'll translate this into an msprime-only MWE and open an msprime discussion.

grahamgower:
Verification plots with 1 million ms (mspms) and msprime replicates. Sims differ very slightly from moments for a couple of the pulse models (`proportion=0.5` in the a/b/c graphs, and @apragsdale suggests the moments approximation might break down for large pulses). But overall, there's excellent agreement.
[verification-1m.pdf](https://github.com/popsim-consortium/demes-python/files/6822547/verification-1m.pdf)

jeromekelleher:
Very impressive! I worry *slightly* about the circularity of testing msprime against msprime, particularly since the ms formatting code is derived from msprime's. But, given we can't pull in ms input it's probably not worth worrying about.

jeromekelleher (approved):

Issue #357, opened 2021-07-09
sort pulses when loading, to avoid ambiguity between discrete-/continuous-time simulators


grahamgower:
See discussion for why we should to do this:
https://github.com/popsim-consortium/demes-spec/issues/99

Probably this can be done in `Graph.from_dict()`.

PR #362, opened 2021-07-23
fix some pulse edge cases

### LINKED ISSUE #328: Graph.assert_close() shouldn't ignore pulse ordering.
The docs currently say that `the order in which admixture pulses were specified` is ignored during the comparison. But this shouldn't strictly be the case. If multiple pulses are specified for the same time, then the order in which pulses were defined will matter. Hence the correct thing to do is to (stable)sort pulses by the time field and then compare.

grahamgower:
Closes #328 and #357.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1617,17 +1617,6 @@ class Graph:
                 "the desired ancestry proportions."
             )
 
-        # Check for multiple pulses into dest at the same time that
-        # give a sum of proportions > 1.
-        proportion_sum = proportion
-        for pulse in self.pulses:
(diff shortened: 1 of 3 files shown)


apragsdale:
We’re moving out of our place today, so probably won’t get to look at this until the weekend or early next week.

apragsdale (approved):
This all looks good, thanks @grahamgower.

I think it makes sense to go ahead and merge this, and then I'll make the needed changes over in #353. There are a few suggestions that I haven't gotten to there, but with this merged I can do that and sync with this all at once.

PR #385, opened 2021-11-28
Closes #384

### LINKED ISSUE #384: Some symmetric migration scenarioes are not simplified
For graphs with continuous symmetric migration between two demes that do not have equivalent time intervals of existence, `graph.asdict_simplified()` does not collapse the directional migrations into a single symmetric migration specified by `demes` and `rate`. Instead, it remains as asymmetric migrations with `source` and `dest`. Those migrations should be collapsed.

For example, using the Gutenkunst-OOA model:

```python
import demes
g = demes.load("examples/gutenkunst_ooa.yml")
print(g)
```
The migrations print as
```
migrations:
- demes: [YRI, CEU]
  rate: 3e-05
- demes: [YRI, CHB]
  rate: 1.9e-05
- demes: [CEU, CHB]
  rate: 9.6e-05
- {source: YRI, dest: OOA, rate: 0.00025}
- {source: OOA, dest: YRI, end_time: 21200.0, rate: 0.00025}
```

The input file has
```
tail -n 5 examples/gutenkunst_ooa.yaml
```
```
migrations:
- {demes: [YRI, OOA], rate: 25e-5}
- {demes: [YRI, CEU], rate: 3e-5}
- {demes: [YRI, CHB], rate: 1.9e-5}
- {demes: [CEU, CHB], rate: 9.6e-5}
```

apragsdale:
Was a pretty quick bug to find - `dest` repeated twice instead of both `source` and `dest` when checking one end of the migration time interval.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -2220,7 +2220,7 @@ class Graph:
                 source = migration["source"]
                 dest = migration["dest"]
                 time_hi = min(self[source].start_time, self[dest].start_time)
-                time_lo = max(self[dest].end_time, self[dest].end_time)
+                time_lo = max(self[source].end_time, self[dest].end_time)
                 if migration["end_time"] == time_lo:
                     del migration["end_time"]
                 if migration["start_time"] == time_hi:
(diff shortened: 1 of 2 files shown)


grahamgower (approved):
LGTM. Nice catch!

PR #386, opened 2021-11-28
Closes #358

### LINKED ISSUE #358: improve support for "Infinity" as a start_time value
We decided to mandate support for the string value `"Infinity"`, regardless of input format (https://github.com/popsim-consortium/demes-spec/pull/101). So we'll need to make sure strings work when using the Builder API. I think also the defaults section has been missed in the current "unstringify_infinities()" function.

apragsdale:
~~Check if "Infinity" is given as a start time, which is converted to `math.inf` in `Builder.add_deme()`. Maybe there is a more appropriate place to handle this?~~ What do you think, @grahamgower 

~~Edit: uses `load_dump._unstringify_infinities()` in `Builder.resolve()`.~~

Edit 2: simply adjust a deme or migration start time in `add_deme()` and `add_migration()`. Using `_unstringify_infinities()` causes some complications where `fromdict()` in `Builder.resolve()` is not able to catch type errors.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -2373,6 +2373,8 @@ class Builder:
         if proportions is not None:
             deme["proportions"] = proportions
         if start_time is not None:
+            if start_time == "Infinity":
+                start_time = math.inf
             deme["start_time"] = start_time
         if epochs is not None:
             deme["epochs"] = epochs
@@ -2427,6 +2429,8 @@ class Builder:
         if dest is not NO_DEFAULT:
             migration["dest"] = dest
         if start_time is not None:
+            if start_time == "Infinity":
+                start_time = math.inf
             migration["start_time"] = start_time
         if end_time is not None:
             migration["end_time"] = end_time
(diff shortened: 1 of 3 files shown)


grahamgower:
What you have here looks right, thanks @apragsdale. In #358, I mentioned that the `defaults` fields are not checked for string infinity value---I guess just `defaults["deme"]["start_time"]`? Maybe there are other fields that could have infinities that I didn't think of when writing _unstringify_infinities()?

Oh, and when you squash, would you be able to change the commit message to indicate what the change is doing (rather than just referring to the issue number)?

apragsdale:
Closes #358.

Ok, updated here. Thanks for the suggestions. I'll wait a bit and then merge unless you think there is something else that is needed here.

> What you have here looks right, thanks @apragsdale. In #358, I mentioned that the `defaults` fields are not checked for string infinity value---I guess just `defaults["deme"]["start_time"]`? Maybe there are other fields that could have infinities that I didn't think of when writing _unstringify_infinities()?

Also added a bit in `_ustringify_infinities()` to check default start times in demes and migrations.

> Oh, and when you squash, would you be able to change the commit message to indicate what the change is doing (rather than just referring to the issue number)?

Good point. Commit message is changed.

Issue #340, opened 2021-06-29
Reject YAML files with values set to `null`.


grahamgower:
See https://github.com/popsim-consortium/demes-spec/issues/76.

PR #387, opened 2021-11-28
First pass at fixing #340

apragsdale:


apragsdale:
Essentially what I've done here is add a function to `load_dump.py` that iterates through the loaded data dict and checks for any `None` values, raising an error if it finds one. There may be a much more elegant way of doing this, and this solution feels a bit hacky. It does seem to work though, and if we're just forbidding `null` entries in the YAML until we make a final decision on what they mean and how they should be handled, this could be a reasonable stopgap.

Thoughts @jeromekelleher @grahamgower @molpopgen?

demes/load_dump.py
```diff
@@ -106,6 +106,31 @@ def _unstringify_infinities(data: MutableMapping[str, Any]) -> None:
...
 
 
+def _no_null_values(data: MutableMapping[str, Any]) -> None:
+    """
+    Checks for any null values in the input data.
+    """
+
+    def check_if_None(key, val):
+        if val is None:
+            raise ValueError(f"{key} cannot have value `null`")
```
grahamgower, line 116: Maybe the error could be:
```suggestion
            raise ValueError(f"{key} must have a non-null value")
```
I think this is slightly less confusing in the case where someone accidentally omits a value (which might be more common than deliberately setting a value to `null`?).

grahamgower (approved):
LGTM. I can't think of a cleaner way to do this, and your approach is at least simple and self contained. Can you update the commit message to indicate the change before you merge this?

demes/load_dump.py
apragsdale: Good point - I'll change that.

PR #389, opened 2021-11-30
Update changelog for 0.2.0

apragsdale:
I'm hoping I didn't miss anything. I looked back through closed PRs dating from the last release from early June.

cc: @grahamgower


jeromekelleher:
LGTM

Do we want to tag 0.2.0 before adding the top-level metadata?

apragsdale:
> Do we want to tag 0.2.0 before adding the top-level metadata?

I'm not sure what the preferred order of operations is here, so I'll let you two make that call.

jeromekelleher:
Be nice to get all the major changes into 0.2.0 if we could. But if we're ready to ship and want to get code out asap there's no problem in more releases. I'm just bringing it up for discussion really

apragsdale:
Oh I misunderstood, I see what you mean. I agree that would be nice to have it all together in this release, but if it takes a little while we might want to just ship this and deal with the metadata in a follow-up release (it won't be a breaking change, at least). I'd vote sooner than later, so we can start on propagating the changes to msprime (hopefully in time for msprime-1.1), moments, fwdpy11, demesdraw, etc..

grahamgower:
I'm inclined to do a release now and just do another release for the metadata support. As @apragsdale says, the downstream packages will each need (small) fixes for the breakages, and it would be nice to get that stuff sorted sooner rather than later.

PR #396, opened 2021-12-06
Fix some f-strings; additional linting.

grahamgower:
---

I noticed a missing "f" on an f-string and wanted to check for other occurences. It turns out there's a lack of tooling around this (for obvious reasons - there would be lots of false positives). However, there is an [open pr for pylint](https://github.com/PyCQA/pylint/pull/4787) adding checks for this, so I pulled that down and ran it. I checked some other stuff too using pylint, but in general its value seems marginal.


jeromekelleher (approved):
LGTM

PR #412, opened 2021-12-09
Create "stable" symlink for the docs.

grahamgower:
---
The idea here is that we can link to a 'stable' docs path from external projects (e.g. for the sphinx objects.inv file). I manually created a stable symlink in the current docs repo.

### DIFF
--- a/docs/is_new_stable.py
+++ b/docs/is_new_stable.py
@@ -1,0 +1,31 @@
+import sys
+
+import packaging.version
+
+
+def is_new_stable(old_stable, tag):
+    """
+    Return 0 if `tag` should be the new 'stable' docs folder, 1 otherwise.
+    """
+    try:
+        v_t = packaging.version.Version(tag)
+    except packaging.version.InvalidVersion:
+        return 1
+    v_s = packaging.version.Version(old_stable)
+    return 0 if not v_t.is_prerelease and v_t > v_s else 1
+
+
+if __name__ == "__main__":
+    assert is_new_stable("0.2.0", "taggy-mac-tag-tag") != 0
+    assert is_new_stable("0.2.0", "0.1.0") != 0
+    assert is_new_stable("0.2.0", "0.1.0b1") != 0
+    assert is_new_stable("0.2.0", "0.2.0b1") != 0
+    assert is_new_stable("0.2.0", "0.3.0b1") != 0
+    assert is_new_stable("0.2.0", "0.2.1") == 0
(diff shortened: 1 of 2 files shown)


docs/is_new_stable.py
```diff
@@ -0,0 +1,31 @@
+import sys
+
+import packaging.version
```
grahamgower, line 3: This is a core infrastructure package, used by setuptools, build, etc. So its almost certainly already installed (but we pip install it just in case).

docs/is_new_stable.py
```diff
@@ -0,0 +1,31 @@
...
+    try:
+        v_t = packaging.version.Version(tag)
+    except packaging.version.InvalidVersion:
+        return 1
+    v_s = packaging.version.Version(old_stable)
+    return 0 if not v_t.is_prerelease and v_t > v_s else 1
+
+
+if __name__ == "__main__":
+    assert is_new_stable("0.2.0", "taggy-mac-tag-tag") != 0
```
apragsdale, line 19: Will these  `"0.2.0"` need to be modified each time we have a new release?

apragsdale (commented):
Looks very useful - just small question about the asserts here.

docs/is_new_stable.py
grahamgower: No, the old_stable is passed in as a command line argument, these are just my unit tests. :) I made this script yesterday and now I'm copying it around to a few repos, so it made sense to have the unit tests in the script. And it makes it self evident what the intended behaviour is in different circumstances.

apragsdale: Ok, gotcha! Was just trying to figure out all that is happening. Learning a lot about github actions/docs/etc from these recent PRs, even though I don't have enough of a knowledge about it to have useful comments.

grahamgower: I'm learning as I go here, but am finally starting to see how the pieces fit together and am also finding it easier to understand why things fail when they fail. I reckon the github actions docs got 50% better since I looked through them earlier in the year. The most pertinent bit is [`GITHUB_REF_NAME`](https://docs.github.com/en/actions/learn-github-actions/environment-variables) which for 'push' events to a branch will be set to the branch name (`main` most of the time for us) and for 'push' events that push a tag it will be set to the tag name. We previously had a sed command that grabbed the ref name from the 'ref', which has a leading 'refs/heads/' for branches and 'refs/tags/' for tags.

jeromekelleher (approved):
LGTM (but not that well up on this stuff)

PR #418, opened 2022-01-01
Don't concatenate different list types.

grahamgower:
Fixes type checking error in #413.

---
Also fix pre-commit (mypy needs attrs in its virtual env), which for some reason had been working for me just fine until today.


jeromekelleher (commented):
LGTM

PR #430, opened 2022-02-18
Test the demes-python implementation against the spec.

grahamgower:
WIP

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -33,6 +33,8 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v3
+        with:
+          submodules: true
 
       - name: setup python
         uses: actions/setup-python@v3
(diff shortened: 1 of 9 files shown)


grahamgower:
There are a whole bunch of failures here, so it will take some time to tidy things up to work on the spec test cases. But I wanted to get feedback on the general approach of including the demes-spec repo as a git submodule. I know submodules are fairly frustrating for folks, but our use case is clear cut and I'm not sure there are any simple(r) alternatives.

Once all the valid test cases can be parsed, we'll also want to confirm that they are resolved identically to the reference implementation. I've just done this for demes-c, so we can do basically the same thing (https://github.com/grahamgower/demes-c/blob/main/tests/compare-reference-implementation.py). This does suggest some duplicated functions, so maybe its worth putting some of this into the demes-spec repo? I guess we could make the reference implementation importable, but this seems like overkill.

jeromekelleher:
Seems like exactly the right approach to me - submodules are bad, but that doesn't mean that they're not sometimes the least-bad way of getting the job done!

I've no problem putting a few simple utility functions into the spec repo. As you say, making it importable is overkill (and not what we want anyway - it's specifically not intended to be used in downstream software).

grahamgower:
This is more or less good to go, pending utf8 decisions and updates to the spec repo.

PR #446, opened 2022-04-25
Bump dependencies; drop Python 3.6.

### LINKED ISSUE #445: drop python 3.6
Various developer dependencies have dropped 3.6 support. The reference implementation parser also requires >= 3.7 and we wish to test against this. It's plausible that we could retain 3.6 support, but require >=3.7 for development, but I'm not interested in dealing with the added maintenance.

The last release works on 3.6, and it's not a bad release at all. So I'm dropping 3.6 for the next demes-python release.

grahamgower:
Closes #445.

### DIFF
--- a/.github/workflows/dev.yaml
+++ b/.github/workflows/dev.yaml
@@ -15,8 +15,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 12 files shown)


jeromekelleher (approved):
LGTM

PR #447, opened 2022-04-28
Update mergify config.

grahamgower:
Also move it under the .github/ folder so its less likely to be
forgotten next time it needs updating.


jeromekelleher (approved):
LGTM. Manually merging as mergify won't do it.

PR #499, opened 2023-03-26
Renaming demes within Graph object

aabiddanda:
I encountered a need for this feature when looking to develop a `snakemake` workflow that expected specific deme labels for post-processing of simulations. The method is fairly straightforward and implements the type-checks. I will update the tests for this as well if it is a useful feature for others to have or if alternative spec-checks should be performed.

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -1997,24 +1997,30 @@ class Graph:
         graph.generation_time = 1
         return graph
 
-    def rename_demes(self, deme_dict={}):
+    def rename_demes(self, rename: MutableMapping[str, str]) -> Graph:
         """
-        Rename demes according to a dictionary
+        Rename demes according to a dictionary.
+
+        :param dict rename:
+            A dictionary with deme names and new names.
+        :return:
+            A demographic model with renamed demes.
+        :rtype: Graph
         """
         graph = copy.deepcopy(self)
-        if not isinstance(deme_dict, dict):
-            raise TypeError("name dict")
-        for k, v in deme_dict.items():
+        if not isinstance(rename, MutableMapping):
+            raise TypeError("rename_dict is not a dictionary")
+        for k, v in rename.items():
             for c in graph.demes:
                 if c.name == k:
-                    c.name = (v,)
+                    c.name = v
                 if c.ancestors is not []:
                     c.ancestors = list(map(lambda x: x.replace(k, v), c.ancestors))
             for m in graph.migrations:
                 if m.source == k:
                     m.source = v
                 if m.dest == k:
-                    m.dest = (v,)
+                    m.dest = v
         return graph
 
     @classmethod
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -5,7 +5,7 @@ import itertools
 import math
 import numbers
 import operator
-from typing import List, Union, Optional, Dict, MutableMapping, Any, Set, Tuple
+from typing import List, Union, Optional, Dict, MutableMapping, Mapping, Any, Set, Tuple
 import warnings
 
 import attr
@@ -1997,9 +1997,9 @@ class Graph:
         graph.generation_time = 1
         return graph
 
-    def rename_demes(self, names: MutableMapping[str, str]) -> Graph:
+    def rename_demes(self, names: Mapping[str, str]) -> Graph:
         """
-        Rename demes according to a dictionary.
+        Rename demes according to a dictionary that may contain a partial set of demes.
 
         :param dict names:
             A dictionary with deme names and new names.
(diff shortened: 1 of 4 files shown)


grahamgower:
Hi @aabiddanda! This is a good idea, so I agree we need to export an API function for this. And I think your choice to put this at `demes.Grap.rename_demes()` is a good one. Actually, there's already an implementation of this function hiding in the ms conversion code (see below). Perhaps you'd like to move the existing code and make the ms converter use the new function instead? The assert should be changed to raise a ValueError instead, and the function probably needs its own set of tests (it looks like I was a bit lazy with the existing tests, and just tested the function that calls the renaming code).


https://github.com/popsim-consortium/demes-python/blob/392c6a0eb5e70223a00d6659df2134317a94bdf0/demes/ms.py#L824-L839

aabiddanda:
@grahamgower great spot on this lurking already in the codebase! I'll move this code to the appropriate place and implement some tests explicitly targeting this functionality to pass `codecov` shortly.

demes/demes.py
```diff
@@ -1997,6 +1997,36 @@ def in_generations(self) -> Graph:
...
+        Rename demes according to a dictionary.
+
+        :param dict names:
+            A dictionary with deme names and new names.
+        :return:
+            A demographic model with renamed demes.
+        :rtype: Graph
+        """
+        graph = copy.deepcopy(self)
+        if not isinstance(names, MutableMapping):
```
grahamgower, line 2011: I don't think its necessary to do explicit type checking here. If a non-dict is passed, this function will fail anyhow with an indexing error of some sort.

demes/demes.py
```diff
@@ -1997,6 +1997,36 @@ def in_generations(self) -> Graph:
         graph.generation_time = 1
         return graph
 
+    def rename_demes(self, names: MutableMapping[str, str]) -> Graph:
```
grahamgower, line 2000: What's the reasoning for this being a `MutableMapping` rather than just a `Mapping`? This function shouldn't be modifying the mapping, right?

grahamgower (changes requested):
Thanks @aabiddanda, this is looking good. Would you be able to expand the tests to include the nominal cases too? I.e. check the deme names after renaming are what the should be, for each of the deme names, ancestors, migration dest/source, pulse dest/source.

demes/demes.py
aabiddanda: The reasoning for the `MutableMapping` was for the scenario where you only want to change the label of a single deme  but not mess with all of the other demes (specifically the ancestral demes). The basis is that if an existing deme name does not exist you may want to default to keeping that name by adding it to the renaming mapping (see code below). 

https://github.com/aabiddanda/demes-python/blob/e2f6eb1fd78dab44045daa080add5f17bdcaf5d9/demes/demes.py#L2013-L2015


This relaxes the requirement that every deme needs to be defined in the names dictionary as well. If you think that adding in conditional statements to the cases to only rename demes that are named (rather than all demes in the graph) I can certainly implement that if its more in keeping with the spec.

grahamgower: Yes, renaming only a subset of the demes seems reasonable, but I don't think we should be mutating the argument here. Using conditionals would be fine, but if you want to keep the code as it is, then you could copy the `names` dict at the start of the function.

demes/demes.py
```diff
@@ -1997,34 +1997,39 @@ def in_generations(self) -> Graph:
...
-                names[deme.name] = deme.name
-        for deme in graph.demes:
-            deme.name = names[deme.name]
-            deme.ancestors = [names[ancestor] for ancestor in deme.ancestors]
+            if deme.name in names:
+                deme.name = names[deme.name]
+                deme.ancestors = [
+                    names[ancestor] if ancestor in names else ancestor
+                    for ancestor in deme.ancestors
+                ]
```
grahamgower, line 2019: I think this misses the case where `deme.name` is not in `names`, but the deme has ancestors that must be renamed.

aabiddanda: Good spot! I think that this should now be fixed in the most recent commit as it passes all of the tests now.

demes/ms.py
```diff
@@ -821,21 +821,11 @@ def migration_matrix_at(time):
     return graph
 
 
-def remap_deme_names(graph: demes.Graph, names: Mapping[str, str]) -> demes.Graph:
+def remap_deme_names(
+    graph: demes.Graph, names: MutableMapping[str, str]
+) -> demes.Graph:
```
grahamgower, line 826: I guess this can stay as a Mapping now too?

aabiddanda:
@grahamgower should the changelog be changed as well? I don't think that this necessarily constitutes a new version per-se but to maintain that it is a new feature in a to-be-released version? Let me know if this would be easier to handle closer to when you are thinking about releasing another version.

grahamgower (approved):
This looks great!

grahamgower:
Don't worry about the changelog - I'll update that when I do the next release.

PR #505, opened 2023-04-01
Fix use of numpy strings for deme names

### LINKED ISSUE #495: Crash on dict conversion if using numpy.str_
MRE using 0.2.2:

```
import numpy as np
from demes import Builder
b = Builder(defaults=dict(epoch=dict(start_size=1)))
b.add_deme(np.str_("a"))
b.resolve().asdict()
# ValueError
```

What's happening: `numpy.str_` defines a `__float__` method, but float conversion throws for non-numeric strings. This causes https://github.com/popsim-consortium/demes-python/blob/1c8a9d9c5f82aa24276f8858ba39a9a9a6be6746/demes/demes.py#L2350-L2351 to fail.

Suggested fix: probably just wrap `coerce_numbers()` in a `try ... except` and return a string if the conversion fails. I have a PR for this if you're interested.

How this arose: During randomized testing. We randomly assign source and dest populations using something like 

```source, dest = np.random.choice(["A", "B", "C", "D"], 2, replace=False)```

This returns a Numpy string even if the input array is pure Python.

grahamgower:
Closes #495

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -2369,21 +2369,27 @@ class Graph:
                 or (not (hasattr(value, "__len__") and len(value) == 0))
             ) and attrib.name != "_deme_map"
 
-        def coerce_numbers(inst, attribute, value):
-            # Explicitly convert numeric types to int or float, so that they
+        def coerce_types(inst, attribute, value):
+            # Explicitly convert numeric and string types, so that they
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM

Issue #576, opened 2024-11-15
Graphs inconsistent w/the specification are accepted


molpopgen:
The Python API will accept this graph:

```python
import demes

yaml = """
time_units: generations
demes:
 - name: A
   epochs:
    - start_size: 1000
 - name: B
   epochs:
    - start_size: 1000
 - name: C
   epochs:
    - start_size: 1000
pulses:
- sources: [A]
  dest: C
  proportions: [0.]
  time: 10
"""

g = demes.loads(yaml)
```

The JSON schema of the [specification](https://popsim-consortium.github.io/demes-spec-docs/main/specification.html) excludes 0 as a valid value for a proportion:

```json
  proportion:
    type: number
    exclusiveMinimum: 0
    maximum: 1
```

Some casually grepping of the test cases in the demes-spec repo indicated that 0.0 doesn't appear anywhere over there.  I could have missed it, though.

grahamgower:
Fixed by #577.

PR #577, opened 2024-11-20
Handle invalid proportions

molpopgen:
See #576

### DIFF
--- a/demes/demes.py
+++ b/demes/demes.py
@@ -52,6 +52,11 @@ def unit_interval(self, attribute, value):
         raise ValueError(f"must have 0 <= {attribute.name} <= 1")
 
 
+def unit_interval_exclusive_lo(self, attribute, value):
+    if not (0 < value <= 1):
+        raise ValueError(f"must have 0 < {attribute.name} <= 1")
+
+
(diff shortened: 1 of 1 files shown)


grahamgower:
It looks like the runner is trying to execute:
```
pip install --upgrade pip pip install -r requirements.txt
```
Which kinda makes sense in light of the error message:
```
ERROR: Could not find a version that satisfies the requirement install (from versions: none)
ERROR: No matching distribution found for install
```
That part of the .yml files hasn't changed, but maybe there's a newline issue here that is newly problematic? Mixed dos/unix newlines?

molpopgen:
@grahamgower -- I think some of the issue has to do w/the CI being quite out of date.  I saw similar errors in the demes-spec PRs before the CI there got updated.

grahamgower:
The docs build seems to be working here. It's also working for me offline, with Python 3.11. Maybe just rebase this and apply fixes for the proportions? We can do the update to Python 3.12 separately.

grahamgower (approved):
LGTM.

grahamgower:
Oooh, codecov isn't broken, it really is just overloaded!

PR #579, opened 2024-11-20
fix CI

molpopgen:


grahamgower:
There are three distinct issues here. I fixed the dev-deps problem over in #580. There's a second issue with moving to python 3.12, relating to a Warning in a test. The third issue is intersphinx.

@molpopgen, do you want to take care of the python 3.12 problem, and I'll take a look at intersphinx?

molpopgen:
> @molpopgen, do you want to take care of the python 3.12 problem, and I'll take a look at intersphinx?

I'll plug away at it some more.  It is a bit of a frustrating one and I need to evaluate the various possible solutions.

molpopgen:
@grahamgower -- I reset this one back to main and then bumped the macos runner and py versions for the tests.  CI in #577 is not starting due to the old macos runner.

molpopgen:
@grahamgower -- other than codecov this seems to be working.  The macos update seems necessary for other PR.

grahamgower:
Ok, thanks. I tried rerunning the tests again, but codecov is still failing to upload. I guess the codecov error is wrong and maybe we need to bump that too.

PR #584, opened 2024-11-26
bump codecov action

molpopgen:


molpopgen:
@grahamgower -- bumping the action to v5 didn't do the trick.  It seems that some kind of authentication needs to happen.  I don't think I have the ability to do that.

grahamgower:
For open source projects, Codecov has an exemption to requiring a token. But then we go into a queue for uploads that is shared with all the other open source projects, and it seems that items are often dropped from the queue. I'll look into what is required for getting a token.

grahamgower:
After adding the 'token:' line in the workflow file, the codecov upload just failed due to auth. Now I've added the token in the repository secrets and I'll rerun the failed tests...

grahamgower:
> @grahamgower -- bumping the action to v5 didn't do the trick. It seems that some kind of authentication needs to happen. I don't think I have the ability to do that.

@molpopgen, you are in the demes-maintainers github group which is an admin for demes-python repo. Codecov inherits its admin settings from github, so you should also have the ability to poke anything on the codecov side (this is a bit opaque at first, but just log in to codecov and your repos/organisations will show up).

PR #590, opened 2026-03-13
Move dependency management to uv

### LINKED ISSUE #589: Moving forward with maintenance
We have a couple of action items that would be nice to knock off the plate.  

First, #577 merged a bug fix but has not been released. We should do a release!

Second, it would be useful to streamline our dependency management prior to a release.  In late 2024 (omg), I made separate "requirements" files for various elements of this project (packages/tests/docs/etc..).  The resulting setup is far from ideal. For example, there are many automatically-opened [security issues](https://github.com/popsim-consortium/demes-python/security) for this repo at the moment.  They all (I think...) come from the docs.  However, the requirements file for the docs is difficult to update piecemeal: we cannot update jupyter-book itself (docs will just break....) and updating the various insecure deps will lead to "dependency hell" because we also need to work out what deps of those deps also need bumping.

I propose that we drag ourselves into 2023 and move to using [uv](https://docs.astral.sh/uv/guides/install-python/) for managing dependencies for the CI and development environments.  I did this for fwdpy11 a while ago and I think it has paid off with less time lost due to "ecosystem drift".  Jerome also moved tskit/msprime/etc., over, and I suspect folks there will be happy?

Third, it would seem that #574, #575 are low-hanging fruit that would be straightforward to get ready before the next release.

Thoughts?  I'm happy to take the lead on uv.

pinging: @grahamgower @apragsdale @jeromekelleher

### LINKED ISSUE #591: Wheel building work flow only runs on release
This package is simple, but wheel building can be fragile.  We should be running all but the uploading steps during normal CI operations.

molpopgen:
- [x] Move contents from setup.[py|cfg] into pyproject.toml
- [x] Update github workflows
- [x] Delete old requirements files

Closes #589
Closes #591

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,9 @@ docs = [
 [tool.setuptools_scm]
 write_to = "demes/_version.py"
 
+[tool.coverage.run]
+patch = ["subprocess"]
+
 # NOTE to future developers
 # mypy can be configured here, and perhaps
 # "should" be.


pyproject.toml
```diff
@@ -1,3 +1,87 @@
...
+    "Development Status :: 4 - Beta",
+    "License :: OSI Approved :: ISC License (ISCL)",
+    "Operating System :: OS Independent",
+    "Intended Audience :: Science/Research",
+    "Programming Language :: Python :: 3",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Bio-Informatics"
+]
+
+requires-python = ">=3.10"
```
grahamgower: This doesn't match the `_setup.cfg` file.

pyproject.toml
```diff
@@ -1,3 +1,87 @@
...
+[project]
+name = "demes"
+authors = [
+    {name = "PopSim Consortium", email =  "krthornt@uci.edu"}
+]
+description = "tools for describing demographic models"
+readme = {file = "README.md", content-type = "text/markdown"}
+license = {text = "ISC"}
+classifiers = [
+    "Development Status :: 4 - Beta",
```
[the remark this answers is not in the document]
grahamgower: Nevertheless, I'd argue demes is now `Development Status :: 5 - Production/Stable` (https://pypi.org/classifiers/).

pyproject.toml
grahamgower: Demes can be installed and used on 3.7, but building the docs on CI needed a more recent version due to jupyterbook annoyances. I don't think that was incorrect. That said, I don't mind if you want to bump things to the latest supported python version.

grahamgower: 3.10 is a perfectly fine minimum version. If anyone really needs an older Python version, then an older version of `demes` will in all likelihood be fine too.

docs/development.md
```diff
@@ -13,35 +13,92 @@ projects. However, additional dependencies are required during development, as
 developers regularly run the test suite, build the documentation, and assess
 whether their code changes conform to style guidelines. The `requirements.txt`
```
grahamgower: The `requirements.txt` file is gone now, right?

docs/development.md
```diff
@@ -13,35 +13,92 @@ projects. However, additional dependencies are required during development, as
...
+python -m pip install -e . --group test --group lint --group docs
 ```
 
+```{warning}
+Setting up a developer environment using `pip` does something very
+different from the method using `uv`!
+With `uv`, the developer is reproducible due to the contents of `uv.lock`.
+With `pip`, the loose dependency pins in `pyproject.toml` are parsed and used
+to guide the resolution of the version numbers of all dependencies.
+```
```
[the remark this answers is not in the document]
molpopgen: Removed.

grahamgower:
I'm guessing that codecov isn't detecting the test coverage in the subprocesses now for some reason.

